PR #1, opened 2011-09-19
make it easier to move to clojure 1.3

pyr:
by removing dependencies on clojure.contrib material


pyr:
The round two commit is a bit more debateable, since it doesn't necessarily help as far as readability is concerned, but defnk didn't make it in clojure.core.incubator, so it will be left behind in clojure 1.3 it seems

pyr:
With this last batch of commits, clojure contrib should be eliminated from storm's dependencies. I only have the simple topologies to play with right now. If more complex use cases don't generate any odd behavior, contrib can be removed from dependencies and people will be able to start playing with 1.3

nathanmarz:
Making Storm Clojure 1.3 compatible is definitely a good thing to do. A few notes on these commits:
1. Like Sam said, some and find-first are not equivalent. We can just copy the find-first implementation in backtype.storm.utils to get rid of the dependency.
2. Likewise, instead of dirtying up the code, how about we just copy defnk into backtype.storm.utils?

Copying those functions isn't great, but unless there's a Clojure 1.3 compatible dependency we can pull in, I think it's the best we can do.

pyr:
Instead of copying find-first i used (first (filter)) which uses the same result. I usually use some with fn's returning the element which yields the same result, sorry.

I fixed the defnk uses, I just feel there's not that much to warrant copying the function over, but if you feel it's too ugly i can copy it.

nathanmarz:
Yes, I'd like to keep the usage of defnk. Let's just copy that into backtype.storm.util. To keep the history clean, I think it would be best to make a new pull request with one commit for the defnk change, one commit the the find-first change, and then other commits for the other minor changes you made. I'd do it myself, but I'd like you to get credit for the commits!

pyr:
OK, will do that

On Mon, Sep 26, 2011 at 9:33 AM, Nathan Marz
reply@reply.github.com
wrote:

pyr:
Done, I'll just need to resort the history now

PR #20, opened 2011-10-10
Add Tuple.__repr__ to Python library

twpayne:
This small patch adds a **repr** method to the Tuple class in the Python library. It's very helpful for debugging bolts :-)

Regards,
Tom


nathanmarz:
Thanks! This is very helpful.

PR #24, opened 2011-10-12
Upgrade Thrift

ptgoetz:
Upgraded to thrift version 0.6.1 to enable compatibility with latest version of Cassandra.


nathanmarz:
Thanks. It looks like there's a few more changes that need to be made regarding the ByteBuffer API change. I'll try to get to this over the weekend.

nathanmarz:
The master branch is now upgraded to Thrift 0.7.0.

PR #26, opened 2011-10-13
fixed possible client config parsing issue

bbarkerfloss:
was having some trouble getting "production" cluster running locally. I was getting an exception in NimbusClient.java. Looks like the YAML parser was giving back an ArrayList and we were casting it to a String. This fixes the issue, but given no one else has run into this I wonder if its because I was doing something else wrong.


bbarkerfloss:
I just realized what I was doing wrong. My YAML is out of date. The nimbus node will always be a singleton.

-Jeff

PR #29, opened 2011-10-16
When developing in local mode when a task fails we don't want the whole process to halt

fbrubacher:
I attempted a small fix to this issue, with my addition the only difference is that the process doesn't halt instead it just fails (which works OK for me because i always shutdown a local-topology after some time, but i need to further patch this to shutdown automatically when an error is thrown).

Can you please provide assistance on how to shutdown everything from within task or within worker/mk-worker so i can pass it down to mk-tasks? I can retry the patch again.

I saw kill-local-storm-cluster in testing.clj but i don't seem to have for example a reference to nimbus.

### DIFF
--- a/src/clj/backtype/storm/daemon/task.clj
+++ b/src/clj/backtype/storm/daemon/task.clj
@@ -185,8 +185,9 @@
                        (.report-task-error storm-cluster-state storm-id task-id error))
         
         report-error-and-die (fn [error]
-                                (report-error error)
-                                (halt-process! 1 "Task died"))
+                               (report-error error)
+                               (when (not= "local" (storm-conf STORM-CLUSTER-MODE))
+                                 (halt-process! 1 "Task died")))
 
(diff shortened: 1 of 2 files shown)


fbrubacher:
Hi Nathan , I did what we commented on last time on irc, please send your input and if needed i will try again . Thanks a lot

nathanmarz:
I just saw that you updated this. I'll try to take a deeper look at this later this week.

fbrubacher:
yeah we can chat during the conj :) would love to help further with other issues :)

nathanmarz:
Well, I implemented this portion of this improvement, but it turns out more work is needed to prevent worker and supervisor from halting process in local mode.

nathanmarz:
I'm going to close this. I think this is too complex to implement.

nathanmarz:
BTW, if there are any expected exceptions in a topology, the best practice is to catch those errors and make appropriate calls to OutputCollector#fail.

PR #36, opened 2011-11-02
Add lettuple macro

twpayne:
This pull request adds a `lettuple` macro which provides a convenient way for unpacking tuples in bolts. It saves you from having to manage indices, and allows you to conveniently specify per-value transformations.

Example:

<pre>
(lettuple tuple [i Integer h (String read-json)]
  (prn i)
  (prn h))
</pre>

expands to:

<pre>
(let [tuple79 tuple
      i       (. tuple79 getInteger 0)
      h       (-> (. tuple79 getString 1) read-json)]
  (prn i)
  (prn h))
</pre>

Note that `tuple79` is a symbol generated by the compiler that will be guaranteed to be unique. The first binding of `tuple79` to `tuple` ensures that `tuple` is only evaluated once.

The syntax is:

<pre>
(lettuple tuple [bindings*] exprs*)
</pre>

`binding`s are of the form `var-symbol type-expr`.

If `type-expr` is not a sequence, it is combined with `"get"` to make a method name, e.g. `Integer` becomes `getInteger`, `DoesntExist` becomes `getDoesntExist`. This method is then invoked on `tuple` with the index of the binding.

If `type-expr` is a sequence, the first element of the sequence is treated as above, and the remaining arguments are threaded with `->`. See the example above.

### DIFF
--- a/src/clj/backtype/storm/clojure.clj
+++ b/src/clj/backtype/storm/clojure.clj
@@ -52,6 +52,24 @@
            [(hint id 'java.lang.Object)]
            )
 
+(defmacro lettuple [tuple bindings & forms]
+  (let [tuplex       (gensym "tuple")
+        getter       (fn [type-expr idx]
+                         `(. ~tuplex ~(->> type-expr (str "get") symbol) ~idx))
+        init-expr    (fn [type-expr idx]
+                         (if (seq? type-expr)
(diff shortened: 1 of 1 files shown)


nathanmarz:
It's a good idea, but the right approach will be to make Tuple implement ISeq so that you can use all of Clojure's sequence operations on it (including destructuring). There's an issue open about this here:

https://github.com/nathanmarz/storm/issues/25

PR #50, opened 2011-11-21
Patch for Issue #2 (pick available port for Zookeeper in Local Mode)

pathouk:
I took a stab at implementing this, and it seems to work for me.  I just changed 1 line and added a utility function.


nathanmarz:
Thanks!

PR #67, opened 2011-12-06
Changed the ShellBolt to use a String[] as its arg so this can be passed to the ProcessBuilder directly...

jatrost:
Changed the ShellBolt to use a String[] as its arg so this can be passed to the ProcessBuilder directly.  This allows you to pass command line param to your ShellBolt.


nathanmarz:
I merged this in. Thanks!

PR #80, opened 2011-12-21
Shell Spout & Python Lib Spout

jimrubenstein:
This time...with less noise.  Sorry for all the confusion!

### DIFF
--- a/src/jvm/backtype/storm/spout/ShellSpout.java
+++ b/src/jvm/backtype/storm/spout/ShellSpout.java
@@ -2,36 +2,176 @@ package backtype.storm.spout;
 
 import backtype.storm.generated.ShellComponent;
 import backtype.storm.task.TopologyContext;
+import backtype.storm.tuple.Tuple;
+import backtype.storm.utils.Utils;
+import java.io.BufferedReader;
+import java.io.DataOutputStream;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
+import org.apache.log4j.Logger;
+import org.json.simple.JSONObject;
+import org.json.simple.JSONValue;
 
 
 public class ShellSpout implements ISpout {
+	
+    public static Logger LOG = Logger.getLogger(ShellSpout.class);
+
+	String _shellCommand;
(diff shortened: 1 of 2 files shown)


nathanmarz:
Finally got a chance to look at this again. I'd like to get this merged in, but there are some issues. I'll just enumerate them:
1. The Python API is wrong. "ack" and "fail" methods only take the id of the tuple, not an actual tuple. Your Java and Python code seem to be inconsistent in that regard.
2. The modifications to "readMsg" are bad (the try... except). If you know you're just going to read a string (and not a json object), you should just use readStringMsg. I would recommend encoding the "next" command just like ack/fail commands by using a map ({"command": "next"}). 
3. You shouldn't assume that the "id" created by nextTuple is a Long. It can be any Object, as shown in the Java API for spouts.
4. Your Java code gets anchors from the multilang process. There's no concept of anchors in spouts. Since those anchors are unused, this seems to be a typo.

I'd like to get this finished soon, so I'll be much faster with my feedback from now on.

jimrubenstein:
Thanks for looking over the changes, I'll work on addressing the things you mentioned and commit them to this pull request.

tomjack:
Hey, I'm working on #85 (tentatively also ShellSpout, as I'm trying to factor out the common code). Couple questions:
- In Jim's patch, there seems to be trouble analogous to #85 (while the subprocess is not emitting tuples, it won't get any acks/fails). How does this sound: ShellSpout spawns a thread that blocks on reading the process output, loading the emitted tuples onto a queue that is polled in nextTuple with a 1ms (?) sleep.
- Nathan, you say that we should not assume the id from nextTuple is a Long. Do you mean we should just directly use any String/JSONObject/JSONArray/Long/Boolean the spout sends over?

nathanmarz:
Thomas, yes on both. nextTuple should most definitely be non-blocking.

jimrubenstein:
Sorry I haven't gotten back around to working on this stuff; I had to move on and take care of some other projects - I hope to get back to it at some point.  @Tomo, thanks for picking it up on this, I will absolutely want to/need to use a ShellSpout to implement the features I have coming up....damn time, there's never enough of it.

nathanmarz:
Closing this because @tomo finished this feature in another pull request

PR #81, opened 2011-12-21
Fixed issue 68

killme2008:
Fixed issue 68: use curator to handle zookeeper.

### DIFF
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -23,6 +23,7 @@ nimbus.reassign: true
 nimbus.file.copy.expiration.secs: 600
 
 ui.port: 8080
+ui.context.path: "/"
 
 drpc.port: 3772
 
--- a/project.clj
+++ b/project.clj
@@ -13,7 +13,7 @@
                  [storm/libthrift7 "0.7.0"]
                  [clj-time "0.3.0"]
                  [log4j/log4j "1.2.16"]
-                 [org.apache.zookeeper/zookeeper "3.3.2"]
+                 [com.netflix.curator/curator-framework "0.6.4"]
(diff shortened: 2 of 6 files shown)


killme2008:
All uni test cases passed successfully.

killme2008:
Add a new option "ui.context.path",it is "/" by default.

I have to access storm ui by a context path such as "http://host:8080/storm-ui",then i can set ui.context.path: "/storm-ui" to get it done.

killme2008:
I've run this patch on my storm cluster for one week now,it seems all is fine.

nathanmarz:
What's the point of the ui context path stuff? This is unrelated to using curator for zookeeper and should be discussed separately.

killme2008:
I am sorry.It seems that github will add all commits to the pull requests.The patch is only 

https://github.com/killme2008/storm/commit/37665fd19df98f66096b36b07eae89fcf1debb86
and:
https://github.com/killme2008/storm/commit/d891c79c6c8ba7ec65f6524f573ee0c6ec60e061

The ui.context.path option is only useful for us,we have a system to manage all system's web console,and it need a conext path for authority and dispatching.

nathanmarz:
OK, I'm close to being able to merge this stuff in. One question: why do you call isStarted before calling close?

killme2008:
When I didn't call isStarted  method before close, some unit test cases failed because of closing curator more than once.Curactor doesn't allow calling close method more than once.And it seems that some unit test cases do it.So i have to check if the curator client is started before closing it.

nathanmarz:
I merged in those two commits with a couple modifications. Thanks for the patch!

PR #100, opened 2012-01-13
Adds storm.zookeeper.retry.times and storm.zookeeper.retry.interval 

killme2008:
This patch adds two new options for zookeeper:

```
storm.zookeeper.retry.times
storm.zookeeper.retry.interval 
```

They define zookeeper clients retry policy.Please check the difference.


nathanmarz:
Merged in with some modifications (to unify the code with the usage of zookeeper in transactional topologies). Thanks!

PR #101, opened 2012-01-15
Fix for Issue #82 (Storm should extract resources dir in local mode when appropriate)

dkincaid:
This should be a fix for issue #82. It checks the url that the supervisor gets from the getResource call for "jar:" on the front to determine if it's inside a jar file. It calls the Java function Utils.copyResourcesFromJar() which extracts the files from the jar and copies them to the given directory.

I didn't do extensive testing, but it worked here for me on a couple of different situations.

### DIFF
--- a/src/clj/backtype/storm/daemon/supervisor.clj
+++ b/src/clj/backtype/storm/daemon/supervisor.clj
@@ -1,7 +1,8 @@
 (ns backtype.storm.daemon.supervisor
   (:use [backtype.storm bootstrap])
   (:use [backtype.storm.daemon common])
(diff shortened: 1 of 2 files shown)


nathanmarz:
It turns out that the URL technique doesn't work in all cases (it doesn't work on my computer), so I took a different approach to fix this.  See https://github.com/nathanmarz/storm/commit/d29cf6a591869989edb42629287420fcadd321c0

The fix is available in master. Try it out and let me know how it goes. If you confirm it works, I'll include this fix in 0.6.2.

PR #105, opened 2012-01-20
Update childopts for supervisor; add childopts for ui; bin/storm polish

ddillinger:
- Changed supervisor() to use supervisor.childopts (looks like just a copy/paste/nochange)
- Created a 'storm help' command that re-uses the existing print_usage
- Made unknown commands default to 'storm help' instead of stacktracing
- Added ui.childopts to ui(), to include JVM args to ui processes
- Added default -Xmx768m to ui.childopts entry in defaults.yaml, based on what had been in bin/storm


nathanmarz:
Thanks!

PR #113, opened 2012-01-28
Added check for null servers list on the setup of the DRPCSpout. Fixes #...

danharvey:
...112

Found this bug whilst I was testing out storm drpc for some projects I'm working on. This should make the error logged more clear if users don't add servers to the drpc list before they run a drpc topology.


nathanmarz:
Thanks Dan! I need you to sign a contributor agreement before I merge this in:

PR #125, opened 2012-02-05
push request for issue: 95 

haitaoyao:
it can be used like this:

./bin/storm -c nimbus.host=XXX -c nimbus.thrift.port=6627

all the simple config values can be overriden except for the complex types like zk addresses, which is an array.


haitaoyao:
executeAsync added according to Issue 62

haitaoyao:
the code has been updated according to your coding format. 

thanks.

nathanmarz:
Three things:
1. Can you split this into a separate pull request for each issue?
2. Are you sure your command line configs code works? This doesn't look right: https://github.com/haitaoyao/storm/commit/8d3e385b8dde74506147d710bcb83b3a2e79b6f9#L0R152
3. Does your command line config code work if the config has spaces in it?

haitaoyao:
1. Can you split this into a separate pull request for each issue?  
   OK , I will separate it into 2 pull request.
2. Are you sure your command line configs code works? 
   I will check it out again.  Maybe I made some mistakes while reformatting the code .Sorry for that.
3. Does your command line config code work if the config has spaces in it?
   No, is there any frequently used config options that have to contain spaces? It only works for the simple config options without spaces.

nathanmarz:
topology.worker.childopts is an example of one you might set that could contain spaces in it. We might as well get this feature right the first time without any issues.

Also, you're going to have to merge master into your branch since the pull request can't be automatically merged.

haitaoyao:
options that contains spaces are supported now.

merged to trunk.

thanks

nathanmarz:
I merged this in with a couple minor modifications. Thanks!

PR #135, opened 2012-02-16
pull request for this issue: https://github.com/nathanmarz/storm/issues/104

xumingming:
pull request for this issue: https://github.com/nathanmarz/storm/issues/104


nathanmarz:
Unfortunately, this doesn't solve the issue. It's valid to put your own keys into the topology config, so the type checking being done here won't help with those. I think the right approach should be to validate that serializing and deserializing the config doesn't change the config. In pseudocode:

config.equals(json_deserialize(json_serialize(config))

xumingming:
sorry, i might be misunderstanding what is a 'invalid config' you mean, can you give an example of invalid config? 

I was thinking these config items are all simple types(int, string, double), it should be the same value when you serialize, deseialize back and forth?

nathanmarz:
Here's a test case that reproduces the issue:

lein repl
(from-json (to-json {"aaa" (backtype.storm.testing.TestWordSpout.) "bbb" 23}))

JSON doesn't know how to deal with "TestWordSpout", but instead of throwing an error it just returns nil. The three options for how to deal with this are:
1. Use a different JSON library that doesn't have this behavior (Jackson is a possibility)
2. Add a check that roundtripping a configuration through json serializer/deserializer returns the same result
3. Do something clever to allow any object to be used as values

I'm leaning towards #1 or #2 right now since for simplicity.

xumingming:
ah, i understand the issue now. your option 1 seems not very applicable since it might need to change the **from-json** method signature, because when you call **from-json** to convert json string back into java bean, you dont have the class information -- so you can not convert a json string back into a java bean without extra class information. If people want to pass in a java bean config, he can use map instead:

``` clojure
{"myconfig" {"config1" true "config2" false}}
```

so i will take the second option.

nathanmarz:
There were still some issues with checking the maps properly for equality, so I implemented this myself. See this commit: https://github.com/nathanmarz/storm/commit/29beadd5983931b230024018bd1dc24183a6a2b5

PR #136, opened 2012-02-17
#89 Show/hide implicit parts of topology in Storm UI

tjun:
this request is for issue #89 .
Add checkbox to hide components that prefixed with "__".
But I don't know where the checkbox should be placed.
Please try it.

### DIFF
--- a/src/clj/backtype/storm/ui/core.clj
+++ b/src/clj/backtype/storm/ui/core.clj
@@ -65,7 +65,7 @@
         type: 'numeric' 
     });
     $(document).ready(function() {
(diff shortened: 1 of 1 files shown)


nathanmarz:
Hiding the rows isn't enough, the key part of this issue is how the stats are aggregated. When the box is checked, you want the numbers from the implicit streams to disappear from the aggregate stats as well. So most of this logic should happen after the stats are retrieved from Nimbus by filtering out the implicit streams prior to aggregation.

nathanmarz:
Closing, but feel free to submit another pull request if you address my comment above.

PR #138, opened 2012-02-22
ShellSpout, async ShellBolt, python implementation

tomjack:
The _readTaskIds/_readTuple trampoline/queue stuff in the Python implementation seems funny.. not sure about that — I am not fluent in Python. Still working on a node.js version where the lack of synchronization around task ids after emits is no problem.

I haven't yet tested reliable spouts. Will do that sometime soon.

I made a small change to the protocol, expecting and "end\n" after the pid as well, to be consistent. I could easily add back in the special case.

### DIFF
--- a/src/clj/backtype/storm/clojure.clj
+++ b/src/clj/backtype/storm/clojure.clj
@@ -229,6 +229,7 @@
 (defalias bolt-spec thrift/mk-bolt-spec)
 (defalias spout-spec thrift/mk-spout-spec)
 (defalias shell-bolt-spec thrift/mk-shell-bolt-spec)
+(defalias shell-spout-spec thrift/mk-shell-spout-spec)
 
 (defn submit-remote-topology [name conf topology]
   (StormSubmitter/submitTopology name conf topology))
--- a/src/clj/backtype/storm/thrift.clj
+++ b/src/clj/backtype/storm/thrift.clj
@@ -157,13 +157,24 @@
     {:obj spout :p parallelism-hint :conf conf}
     ))
 
+(defn- shell-component-params [command script-or-output-spec kwargs]
+  (if (string? script-or-output-spec)
+    [(into-array String [command script-or-output-spec])
+     (first kwargs)
+     (rest kwargs)]
+    [(into-array String command)
+     script-or-output-spec
+     kwargs]))
+
 (defnk mk-bolt-spec [inputs bolt :parallelism-hint nil :p nil :conf nil]
   (let [parallelism-hint (if p p parallelism-hint)]
     {:obj bolt :inputs inputs :p parallelism-hint :conf conf}
     ))
 
-(defn mk-shell-bolt-spec [inputs command script output-spec & kwargs]
-  (apply mk-bolt-spec inputs (RichShellBolt. command script (mk-output-spec output-spec)) kwargs))
+(defn mk-shell-bolt-spec [inputs command script-or-output-spec & kwargs]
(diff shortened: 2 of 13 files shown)


nathanmarz:
Awesome, thanks. Before I take a look at this can you sign a contributor agreement? Details here: https://github.com/nathanmarz/storm/wiki/Contributing-to-Storm

tomjack:
I did some very rough performance tests of Java TestWordSpout vs. Python tester_spout.py vs. my node.js tester_spout.js (all with no sleeps). Not sure about these numbers, but it seems like the Java gets 7x the throughput of Python and 6.5x that of node.js.

tomjack:
What should ShellBolt do if the process emits tuples anchored to a tuple that it has already acked/failed?

Die? Emit unanchored and warn? Skip and warn?

nathanmarz:
If you can detect it, you should throw an exception. Otherwise it's ok to leave the behavior undefined.

nathanmarz:
I took a look through this today. Overall, it's pretty good. Two things:
1. You should be able to emit tuples from the ack/fail methods of a spout. So the same protocol as in nextTuple should be in ack/fail as well.
2. What's going on with the recursion in readTaskIds and readTuple in the python code?

tomjack:
Mind if I rebase onto master and force push here?

I want to remove the "add source field info to JSONified TopologyContext" commit, and my latest work is on a branch rebased onto 0.7.0 with some cleanup.

nathanmarz:
That's fine.

tomjack:
Is it permitted to emit inside a spout's open method? Should trying this blow up the task?

nathanmarz:
Yes, it's permitted.

tomjack:
I just pushed my latest.

I changed the recursion weirdness in storm.py to iteration, hopefully that should be clearer. Maybe it could be made simpler.. I haven't thought about it too much.

Reliable spouts now actually work, and you can emit during ack/fail. I updated tester_{bolt,spout}.py to demonstrate this, though I'm sure there are some combinations of arguments I haven't tried yet.

nathanmarz:
Great work, I've merged this in. One question: what does "implements JSONAware" do? 

Now it would be great to update the Ruby adapter to be compatible :)

tomjack:
JSONAware just has a toJSONString method, which TopologyContext already had implemented, so it seemed natural. I believe this is necessary for JSONValue.toJSONString to call the toJSONString method, which allows writeObject in ShellProcess to write TopologyContexts just like the other supported types.

I can take a crack at the Ruby adapter sometime. I was wondering, should storm.{py,rb} be changed to use the standard names for open/prepare? It looks like storm.rb is actually using the constructor 'initialize' (which appears to be useless?), while storm.py uses methods named 'initialize'.

nathanmarz:
Looks like that's an oversight in the ruby code. It should really be calling initialize in "run" like what the python code does.

In the interest of keeping this stuff backwards compatible with people's code, I don't think we should change the names.

tomjack:
I think that's a bit of a problem for the Ruby implementation... `initialize` is the name of the constructor, but the component already needs to be constructed when `run` calls the setup method. So people will have to write `conf = nil, context = nil` and check whether this is the no-arg constructor call or the actual setup.

Keep 'initialize' in the Python, but change to 'open'/'prepare' in the Ruby, since no one could possibly have been using 'initialize' in Ruby anyway? Or maybe there's some Ruby magic workaround.

nathanmarz:
Ah good point... we should definitely rename the ruby method.

PR #142, opened 2012-02-25
Allow symlinking of the storm binary

gabrielgrant:
right now symlinking the bin onto PATH doesn't work properly, because abspath doesn't follow the symlink.


nathanmarz:
Thanks Gabriel! Can you sign a contributor agreement so that I can merge this in? More info here:

PR #143, opened 2012-02-27
add `:debug true` shortcut for components in clojure dsl

tomjack:
A minor convenience...


nathanmarz:
I decided to close this, because I think it creates confusion as to what happens when both :conf and :debug are set. I'd rather keep the semantics clear, and I don't think {TOPOLOGY-DEBUG true} is that bad (you could make a "debug-conf" variable that contains that map, e.g. ":conf debug-conf"

PR #147, opened 2012-03-02
The count of a seq is 1 + the number of remaining seqs through next()

travisfw:
If the seq is partially consumed it shouldn't return the count of the full list of fields. See ASeq.count()


nathanmarz:
Thanks Travis. I know this is a tiny patch, but I need a contributor agreement for all contributions. It should only take you a couple minutes, instructions here: https://github.com/nathanmarz/storm/wiki/Contributing-to-Storm

travisfw:
Sent via hellofax

PR #148, opened 2012-03-04
allow destructuring of tuple arg in defbolt

tomjack:


nathanmarz:
Hmm, I don't think I like this. It would be nice if it could be destructured either as a map or as a list, e.g.

[[a b c]]

or

[{a "a" b "b" c "c"}]

tomjack:
It can — I only added a test for map destructuring, since this proves to me that it's just dropping down to Clojure destructuring, and I don't want to test every feature of that. Could add a list test case too, though.

I'm wondering now, though — if destructuring is allowed, should it happen in the bolt macro instead? This would require adding a special case there for 'execute which adds in the let wrapper. I noticed this when I went to add the list test case to lalala-bolt2.

nathanmarz:
I don't think it should happen in the macro. It should just leverage Clojure's usual destructuring facilities, so that you can destructure the tuple at any point.

tomjack:
By "happen in the bolt macro instead" I mean just that the code in my patch would move from defbolt to bolt, so that destructuring would work for both. It would still just use a let wrapper to take advantage of Clojure's built-in destructuring. This would already work, except reify doesn't support destructuring, which is why a let wrapper is needed.

A bolt writer can of course just do the destructuring manually inside the bolt definition, without this patch:

``` clojure
(defbolt lalala-bolt1 ["word"] [tuple collector]
  (let [[word] tuple
        ret (str word "lalala")]
    (emit-bolt! collector [ret] :anchor tuple)
    (ack! collector tuple)))
```

Is that what you mean?

schleyfox:
I've been using regular destructuring with tuples in clojure successfully for a while now. I'm reasonably sure that you could even do

``` clojure
(defbolt lalala-bolt1 ["word"] [[word :as tuple] collector]
...
```

tomjack:
No, that throws an IllegalArgumentException since reify doesn't support destructuring. Hence this patch.

schleyfox:
Ah, I realized that everywhere I use it, I have an additional macro wrapping it to add some other behavior to the bolt.

nathanmarz:
I looked into this more, and the reason destructuring was blowing up before was because of the type hinting the dsl was automatically adding. That stuff is completely unecessary (not sure why I did that in the first place), so I removed it and destructuring now works fine.

tomjack:
Ah, so reify supports destructuring after all? No wonder I didn't find any complaints.

PR #152, opened 2012-03-07
for issue #145

ghost:


nathanmarz:
Thanks Kasper! Before I can merge this in, I need you to sign a contributor agreement. It should only take a couple minutes.  Instructions here: https://github.com/nathanmarz/storm/wiki/Contributing-to-Storm

ghost:
I will send you tomorrow :)

right now I am working on issue #111. I only need the cleanup of folders, then I am done.

ghost:
added files for issue #111 as well

nathanmarz:
I merged in the first commit.

Your patch for #111 has a few issues, but I think it will be good to go with just a little more work. First off, the task dirs should be kept under the worker root, not under the new local dir you made ("/tasks"). This is because in certain situations, it's possible for multiple instances of the same worker to spawn (before the supervisor shuts one of them down), so you don't want them trampling on each other's state. That's why each worker is given a unique directory. If you keep the task dirs under the worker dir, then the existing cleanup code will work.

Another issue is that you only need to expose the directory for "this task" in the TopologyContext. No reason to expose the full mapping from task -> dir.

BTW, in the future, be sure to send patches for separate issues as separate pull requests. It makes it a lot easier for me to manage these things.

ghost:
Hi Nathan
Sure I will improve on the patch :)

The reason I didn't put the task dirs under the worker root, was that I
observed new folders were added to them, such as a heartbeat folder, which
I assumed was going to make trouble :O ... This is no problem?

I will change it, so we only get the directory for "this task" :)

I actually tried to make the pull request in a new "thread" on github, but
couldn't figure it out. To be honest, I only have very limited experience
with git, and my experience with github extends to cloning repositories.

Should I make a new patch to improve on the old one, or delete the old
patch and start for scratch?

Kasper

Den 12. mar. 2012 03.59 skrev Nathan Marz <
reply@reply.github.com

nathanmarz:
Right, that's why the task dirs should be in a subdirectory under the worker dir, e.g.:

{root}/workers/{worker id}/tasks/{task id}

The right way to do pull requests is to make a separate branch for every request, but you can just do that for next time :)

You can start a new patch or improve this one, whatever is easier for you. I'll end up merging the aggregate so it makes no difference to me.

PR #157, opened 2012-03-18
prevent duplicates on Fields

git2samus:
Fields receives a list of strings that is stored internally on an ArrayList however it is indexed with a HashMap and field indexes can be retrieved by name (Tuple fields can be retrieved by name too) therefore Fields should prevent duplicates to avoid ambiguity and internal inconsistencies between _index and _fields.

I couldn't find tests for this class so I've added those as well.

### DIFF
--- a/src/jvm/backtype/storm/tuple/Fields.java
+++ b/src/jvm/backtype/storm/tuple/Fields.java
@@ -17,7 +17,14 @@ public class Fields implements Iterable<String>, Serializable {
     }
     
     public Fields(List<String> fields) {
-        _fields = new ArrayList<String>(fields);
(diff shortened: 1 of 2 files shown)


nathanmarz:
This is great. Can you sign a contributor agreement so that I can merge this in? It should only take a couple minutes. Instructions here: https://github.com/nathanmarz/storm/wiki/Contributing-to-Storm

git2samus:
sent

nathanmarz:
Merged, thanks!

MooreTattz (commented):
test/clj/backtype/storm/fields_test.clj

PR #161, opened 2012-03-21
Simplify shell protocol

tomjack:
Just uses JSON for everything, including the pid/piddir, which cleans up both the Java implementation and multilang implementations slightly.

This is currently based on your fix-shell-encoding.

### DIFF
--- a/src/jvm/backtype/storm/spout/ShellSpout.java
+++ b/src/jvm/backtype/storm/spout/ShellSpout.java
@@ -33,7 +33,7 @@ public class ShellSpout implements ISpout {
         _collector = collector;
 
         try {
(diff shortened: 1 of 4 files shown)


tomjack:
Didn't notice the integration tests commented out. I will rebase when your fixes are done.

This also doesn't use JSON for the sync messages. That winds up being clean in multilang implementations anyway, and no reason to add an extra JSON encode to every spout emit.

The protocol _could_ really be JSON everywhere if we made the sync and end messages JSON too. Then a multilang implementor who doesn't want extra JSON encodes everywhere just writes `print '"sync"'`. But this seems silly.

tomjack:
Rebased.

PR #165, opened 2012-03-29
JSON compliance in shell protocol, with Python and Ruby impls

tomjack:
I changed the protocol like so:

```
-> {"pidDir": "...", "conf": {...}, "context": {...}}
<- {"pid": 123}
-> {"command": "next"}
<- {"tuple":["bertels人"],"id":1,"command":"emit"}
-> [3]
<- {"command":"sync"}
```


nathanmarz:
Merged. Great work.

PR #167, opened 2012-03-30
replace slf4j 1.5.8 with slf4j 1.6.4

markhamstra:
Without the changes, slf4j-log4j12 resolves to version 1.5.8, which ends up conflicting with slf4j-api 1.6 resolved by sbt in a build of ScalaStorm:

`[error] SLF4J: The requested version 1.5.8 by your slf4j binding is not compatible with [1.6]`
`[error] SLF4J: See http://www.slf4j.org/codes.html#version_mismatch for further details.`

With the changes, Storm uses the current slf4j 1.6.4, and everyone seems happy.


nathanmarz:
This is implemented as part of the pull request that reworks the logging system to use logback, so once that's merged in this will be available.

PR #177, opened 2012-04-06
resolved the issue that when a topology has not been assigned a task, the storm-ui breaks

xumingming:
for issue #92


nathanmarz:
I tried this out and it didn't fix the problem.

xumingming:
the page still breaks? i did tested and the page didnt break, and sections: **Topology stats**, **Spouts (All time)**, **Bolts (All time)** remains empty.

xumingming:
Nathan, the screenshot below is how a no-task-assigned-topology looks like in storm UI after my fixing, it is what we expected, right? 

storm-ui screenshot

nathanmarz:
Yea actually I tested it wrong. I'll test again tomorrow and then merge.

nathanmarz:
Yea this works fine :)

PR #178, opened 2012-04-09
add warnning message when not enough slots available for topology

xumingming:
for issue #170


nathanmarz:
Closing this, since I don't think this is really that important.

PR #181, opened 2012-04-09
ensure each process use its unique zk port in local mode

xumingming:
for issue #109 , changed much code in **cluster_test.clj**, because added a **let** clause in each test method, need to format the existing code a little bit.


xumingming:
redone this in pull request #187, so close this.

PR #183, opened 2012-04-11
log warnning message when un-acked tuples in a bolt gets too large

xumingming:
for issue #162 . I added a new config: **topology.max.blot.pending** to let user specify the threshold.


nathanmarz:
Closing this because the map no longer exists anymore in 0.8.0, so this isn't even an issue anymore (metadata about ack vals / anchors are now kept with the tuples, not in an external map, so the state will be garbage collected normally).

PR #184, opened 2012-04-11
redone for issue #111

ghost:


nathanmarz:
This has gotten pretty old, and I think we need to think about this feature more.

PR #187, opened 2012-04-12
ensure each process use its unique zk port in local mode(perfect solution)

xumingming:
perfect solution for issue #109 :)


xumingming:
Since we [cannt loop/recur in try/catch](http://stackoverflow.com/questions/1879885/clojure-how-to-to-recur-upon-exception), the code([function mk-inprocess-zookeeper](https://github.com/nathanmarz/storm/pull/187/files#L1L131)) seems a little ugly.

nathanmarz:
Awesome work! Keep the great patches coming :)

PR #195, opened 2012-04-17
fix issue with shell bolts not restarting

tomjack:
Since exceptions (e.g. when the shell process crashes) are thrown in new threads now, the ShellBolt keeps on running.

Does this patch fix all such issues?

### DIFF
--- a/src/jvm/backtype/storm/task/ShellBolt.java
+++ b/src/jvm/backtype/storm/task/ShellBolt.java
@@ -52,6 +52,7 @@ public class ShellBolt implements IBolt {
     private String[] _command;
     private ShellProcess _process;
     private volatile boolean _running = true;
+    private volatile Exception _exception;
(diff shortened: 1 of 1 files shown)


nathanmarz:
Good catch. I tweaked it a little, mainly generalizing it to set _exception for any Throwable, not just IOException

hellp:
Thanks for the quick fix and merge. I was just testing it on our system. Works fine :)

tomjack:
I considered changing it to Throwable, but wouldn't this inappropriately catch e.g. OOM errors? Is that not a problem because the Error will just get thrown again at the beginning of execute?

Also I'll note that hellp alerted me to this problem, I only quickly provided the fix :)

nathanmarz:
Yea, it's fine because it will just get rethrown. It's important that any problem with those threads causes the worker to crash.

PR #197, opened 2012-04-18
multilang: Use new-style classes in Python.

hellp:
To enable proper subclassing including super() usage, new-style classes are required.


nathanmarz:
Merged, thanks!

PR #200, opened 2012-04-20
Optimize sending tuples between tasks colocated in same worker

xumingming:
for issue #172

I changed the messaging protocol a little:
- the connection no longer listen on virtual-port(task-id), they listen on a worker port now.
- send-local-task-empty function is removed  -- because the task does not listen on a socket anymore, we dont need to unblock the socket, we can just put an empty byte-array into the receive-queue directly.


src/clj/backtype/storm/daemon/worker.clj
```diff
@@ -53,11 +53,10 @@
             (-> (reverse-map task->component) (select-keys components) vals)))
     ))
 
-(defn mk-transfer-fn [storm-conf context transfer-queue]
-  (let [^KryoTupleSerializer serializer (KryoTupleSerializer. storm-conf context)]
-    (fn [task ^Tuple tuple]
-      (.put ^LinkedBlockingQueue transfer-queue [task (.serialize serializer tuple)])
-      )))
+(defn mk-transfer-fn [transfer-queue]
+  (fn [task ^Tuple tuple]
```
nathanmarz, line 52: Could we put the logic for sending directly to a local task's receive queue here? A single send thread isn't needed for sending to a local task, so we might as well avoid it here.

src/clj/backtype/storm/daemon/worker.clj
```diff
@@ -118,15 +118,19 @@
...
+        transfer-fn (mk-transfer-fn transfer-queue)
         refresh-connections (fn this
                               ([]
                                 (this (fn [& ignored] (.add event-manager this))))
                               ([callback]
                                 (let [assignment (.assignment-info storm-cluster-state storm-id callback)
                                       my-assignment (select-keys (:task->node+port assignment) outbound-tasks)
-                                      needed-connections (set (vals my-assignment))
+                                      ;; we dont need a connection for the local tasks anymore
+                                      needed-connections (set (vals (filter #(not (contains? task-ids-set (key %))) my-assignment)))
```
nathanmarz: A cleaner way to write this is:

(->> my-assignment
        (filter #(->> % key (contains? task-ids-set) not))
        vals
        set)

src/clj/backtype/storm/daemon/worker.clj
```diff
@@ -175,7 +179,7 @@
                             )
                           :priority Thread/MAX_PRIORITY)
         suicide-fn (mk-suicide-fn conf active)
-        tasks (dofor [tid task-ids] (task/mk-task conf storm-conf (mk-topology-context tid) (mk-user-context tid) storm-id mq-context cluster-state storm-active-atom transfer-fn suicide-fn))
+        tasks (dofor [tid task-ids] (task/mk-task conf storm-conf (mk-topology-context tid) (mk-user-context tid) storm-id mq-context cluster-state storm-active-atom transfer-fn suicide-fn (receive-queue-map tid)))
```
nathanmarz: Can we get rid of the mq-context here? Tasks don't need it anymore.

src/clj/backtype/storm/daemon/worker.clj
```diff
@@ -190,24 +194,42 @@
...
                       (let [node+port->socket @node+port->socket
                             task->node+port @task->node+port]
-                        (doseq [[task ser-tuple] drainer]
-                          (let [socket (node+port->socket (task->node+port task))]
-                            (msg/send socket task ser-tuple)
-                            ))
-                        ))
+                        (doseq [[task tuple] drainer]
+                          ;; if its a local-task, add the tuple to its receive-queue directly
+                          ;; otherwise, send it through the socket.
```
nathanmarz: See note above about moving this to the transfer-fn to be more efficient. A side effect of that is you could move serialization there as well (so the only messages that go onto the transfer queue are ones going external to this process). This would parallelize the serialization.

src/clj/zilch/virtual_port.clj
```diff
@@ -25,21 +27,6 @@
 (defn virtual-url [port]
   (str "inproc://" port))
 
-(defn- get-virtual-socket! [context mapping-atom port]
-  (when-not (contains? @mapping-atom port)
```
nathanmarz, line 29: It seems like we should be able to simplify the virtual port stuff a lot, at the very least unify the regular and "fake" virtual ports. The only difference between them is what kind of connection they bind to: zeromq or local.

src/clj/zilch/virtual_port.clj
```diff
@@ -85,6 +73,35 @@
...
         ))))
 
+(defn launch-fake-virtual-port! [context storm-id port receive-queue-map deserializer]
+  (let [socket (-> context (msg/bind storm-id port))
+        vthread (async-loop
+                 (fn []
+                   (let [[port ser-msg] (msg/recv socket)
+                         msg (if (nil? ser-msg)
+                               nil
+                               (.deserialize deserializer ser-msg))
```
nathanmarz: Virtual port shouldn't do deserialization. It should just put the serialized message onto the receive queue for the task. Let the task handle deserialization (since it is already) – this allows deserialization to happen in parallel.

src/clj/backtype/storm/daemon/task.clj
```diff
@@ -379,10 +380,15 @@
              (Time/sleep 100)))
          ))
      (fn []
-       (let [^bytes ser-msg (msg/recv puller)]
+       (let [msg (.take receive-queue)
```
nathanmarz: This code should be pulled out into its own function since its used in both spout and bolt code paths.

src/clj/backtype/storm/daemon/task.clj
```diff
@@ -291,7 +292,7 @@
     (stats/spout-acked-tuple! task-stats (:stream tuple-info) time-delta)
     ))
 
-(defmethod mk-executors ISpout [^ISpout spout storm-conf puller tasks-fn transfer-fn storm-active-atom
+(defmethod mk-executors ISpout [^ISpout spout storm-conf receive-queue tasks-fn transfer-fn storm-active-atom
```
nathanmarz: Need to type hint receive-queue to avoid reflection.

src/clj/backtype/storm/daemon/task.clj
```diff
@@ -406,7 +412,7 @@
     ;; TODO: this portion is not thread safe (multiple threads updating same value at same time)
     (.put pending key (bit-xor curr id))))
 
-(defmethod mk-executors IBolt [^IBolt bolt storm-conf puller tasks-fn transfer-fn storm-active-atom
+(defmethod mk-executors IBolt [^IBolt bolt storm-conf receive-queue tasks-fn transfer-fn storm-active-atom
```
nathanmarz: Need to type-hint the receive-queue

src/clj/backtype/storm/daemon/worker.clj
```diff
@@ -190,24 +200,37 @@
...
-        virtual-port-shutdown (when (local-mode-zmq? conf)
-                                (log-message "Launching virtual port for " supervisor-id ":" port)
-                                (msg-loader/launch-virtual-port!
-                                 (= (conf STORM-CLUSTER-MODE) "local")
-                                 mq-context
-                                 port
-                                 :kill-fn (fn [t]
-                                            (halt-process! 11))
-                                 :valid-ports task-ids))
+        deserializer (KryoTupleDeserializer. storm-conf (mk-topology-context nil))
```
nathanmarz, line 207: As per my comment below, we should be able to get rid of the special case of "local-mode-zmq?" here. Both local and remote should use a virtual port, the only difference being the kind of port bound to by the virtual port function.

nathanmarz:
Overall, really good job! This is a really exciting improvement. I left comments on the pull request for things that need to be addressed.

xumingming:
I think virtual-port is not a very appropriate name for what it is actually doing now, and because what it is doing is: receive the messages from external tasks, so i give it a new name: receive-thread, hope it make sense to you. And also the logic of receive-thread is totoally moved from virtual-port.clj to loader.clj, so virtual-port.clj is deleted. The received thread logic is unified, only differences is what socket it receive messages on -- LocalConnection or ZMQConnection.

xumingming:
Thanks for all the code review comments, Nathan. Learned a lot.

nathanmarz:
Made some more comments on your last commit. The main issue to address now is one of my last comments about exercising ZMQ in the unit tests (the reason for the STORM_LOCAL_MODE_ZMQ option). 

I think we're really close to having this one finished.

xumingming:
Just found another issue: when use zmq in local mode, _sometimes_ [these lines of code](https://github.com/xumingming/storm/blob/issue172/src/clj/backtype/storm/messaging/loader.clj#L47-L51) cannot kill the receive thread...

nathanmarz:
I fixed the hang that would happen with the receive thread and merged this in. Great work! This is a big improvement.

PR #207, opened 2012-05-06
change to do worker-heartbeat rather than task-heartbeat

xumingming:
for issue #97

### DIFF
--- a/src/clj/backtype/storm/cluster.clj
+++ b/src/clj/backtype/storm/cluster.clj
@@ -95,7 +95,7 @@
   (task-storms [this])
   (task-ids [this storm-id])
   (task-info [this storm-id task-id])
-  (task-heartbeat [this storm-id task-id]) ;; returns nil if doesn't exist
+  (taskbeats [this storm-id])
   (supervisors [this callback])
   (supervisor-info [this supervisor-id])  ;; returns nil if doesn't exist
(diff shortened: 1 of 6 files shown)


xumingming:
I removed the **task-heartbeat** function from **StormClusterState**, because after change to worker-heartbeat, i dont see there is an efficient way to implement this method, I added a **taskbeats** function to return all the taskbeats of a topology instead.

nathanmarz:
backtype.storm.nimbus-test needs to be updated to use the new heartbeat structure. Looks great otherwise.

nathanmarz:
(btw, you may not have seen the error in the tests because with-local-cluster was accidentally suppressing errors. This is now fixed in master)

xumingming:
ah, sorry, i missed that...

nathanmarz:
This is great work! I had to make one critical change to the implementation:

https://github.com/nathanmarz/storm/commit/1df18c318cc9a0c1a739407c5559c9954d71b139#L1R232

The problem is that over time if a topology moves around the cluster, it's possible for Nimbus to use heartbeats from workers that are long dead. You can't just take the heartbeat with the greatest timestamp because the clocks might be skewed. So I changed the code to only read worker heartbeats for assigned tasks, which fixes the problem.

PR #225, opened 2012-05-18
Update to Curator 1.1.3/ZooKeeper 3.4.x

ashleywbrown:
Replaces the Curator 1.0.x series dependency with 1.1.3, which introduces ZooKeeper 3.4.2. Changed the in-process ZK server to use the new NIOServerCnxnFactory in ZooKeeper 3.4.x.

ZK 3.4.x clients are compatible with ZK 3.3.x ensembles.

This change makes Storm compatible with HBase 0.92, which requires the new ZK 3.4 API.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -12,7 +12,7 @@
                  [storm/libthrift7 "0.7.0"]
                  [clj-time "0.4.1"]
                  [log4j/log4j "1.2.16"]
-                 [com.netflix.curator/curator-framework "1.0.1"]
+                 [com.netflix.curator/curator-framework "1.1.3"]
                  [backtype/jzmq "2.1.0"]
                  [com.googlecode.json-simple/json-simple "1.1"]
                  [com.googlecode/kryo "1.04"]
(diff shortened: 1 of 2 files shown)


nathanmarz:
Can you send me a contributor agreement so I can merge this in? Instructions at the bottom of this page (should only take a couple minutes): https://github.com/nathanmarz/storm/wiki/Contributing-to-Storm

ashleywbrown:
Yep, with you shortly.

On 18 May 2012 20:02, Nathan Marz <
reply@reply.github.com

nathanmarz:
Merged, thanks!

nathanmarz:
@ashleywbrown I had to revert this because the new Curator client has a major performance regression. The startup of the client takes a considerable amount of time, making the Storm tests take way longer.

ashleywbrown:
No problem. Hopefully they'll fix it soon!

tomdz:
Any chance that only the Zookeeper portion of this could be applied ? Also, curator is at 1.2.5 now, maybe they fixed the performance regression ?

jasonjckn:
I attempted to upgrade storm to zk 3.4.5, and it caused a unit test regression transactional-test non-determinstically hangs.

PR #232, opened 2012-05-24
Add ruby namespace to thrift file.

azymnis:


azymnis:
Just realized that my editor also removed some whitespace.

nathanmarz:
Closing since it's better that people just generate their Thrift bindings on their own then having Storm generate every possible set of bindings.

PR #236, opened 2012-05-29
a bug fix for CoordinatedBolt

zongyuanwu:
I have seen your fix version and  think it's too complicated.
I make a simple fix , please have a look. Thank you!


zongyuanwu:
I add a map to record  the received tuple count from each source in TrackingInfo. when the CoordinatedBolt received a coordinated-stream-id tuple, it will check the expectedCount with the recored receivedCount,if they are not equal then fail this batch directly.

nathanmarz:
I can't tell what changed because of all the formatting changes. You'll need to use the same formatting as the rest of the Storm codebase for me to review this.

zongyuanwu:
my code is based on your previous version, so there maybe a lot of difference between this code and your current version.

PR #237, opened 2012-05-30
a simpler bug fix version  for CoordinatedBolt

zongyuanwu:
I think this is simpler bug-fix version than yours.
in my code, I make the  CoordinatedBolt ack the tuple from COORDINATED_STREAM_ID stream only if its received tuple count is equal to expectedTupleCount for the specified source task.


nathanmarz:
This stuff is being rewritten to support more general purposes, so this isn't relevant anymore.

zongyuanwu:
ok

PR #238, opened 2012-05-30
Use exec instead of forked child process for JVM

ept:
Using os.system() for creating the JVM forks a child of the Python
process. This makes it difficult for process managers like daemontools
or runit to kill/restart the process at admins' command: the process
manager tracks the PID of the Python process instead of that of the JVM;
and terminating the Python process unfortunately doesn't kill its JVM
child.

Some process managers let you configure the number of times the service
forks, or try to detect forks, but it's a bit of a rabbit-hole. Better
to make the service not fork in the first place.

For that reason I've replaced the call to os.system() with os.execvp(),
which replaces the Python process with the JVM, and thus lets the
process manager track the right PID.

os.execvp() requires that you give command-line arguments as an array of
strings, rather than one long whitespace-separated string, so I had to
change some of the surrounding code accordingly. Note this has the
pleasant side-effect that Storm should now also run if you install it in
a directory that has spaces in its path name.

### DIFF
--- a/bin/storm
+++ b/bin/storm
@@ -82,12 +82,15 @@ def print_remoteconfvalue(name):
     """
     print name + ": " + confvalue(name, [STORM_DIR + "/conf"])
 
-def exec_storm_class(klass, jvmtype="-server", childopts="", extrajars=[], args=[]):
-    nativepath = confvalue("java.library.path", extrajars)
-    args_str = " ".join(map(lambda s: "\"" + s + "\"", args))
-    command = "java " + jvmtype + " -Dstorm.home=" + STORM_DIR + " " + get_config_opts() + " -Djava.library.path=" + nativepath + " " + childopts + " -cp " + get_classpath(extrajars) + " " + klass + " " + args_str
-    print "Running: " + command    
-    os.system(command)
+def exec_storm_class(klass, jvmtype="-server", jvmopts=[], extrajars=[], args=[]):
+    all_args = [
+        "java", jvmtype, get_config_opts(),
+        "-Dstorm.home=" + STORM_DIR, 
+        "-Djava.library.path=" + confvalue("java.library.path", extrajars),
+        "-cp", get_classpath(extrajars),
+    ] + jvmopts + [klass] + args
+    print "Running: " + " ".join(all_args)
+    os.execvp("java", all_args)
 
 def jar(jarfile, klass, *args):
     """Syntax: [storm jar topology-jar-path class ...]
@@ -103,7 +106,7 @@ def jar(jarfile, klass, *args):
         jvmtype="-client",
         extrajars=[jarfile, CONF_DIR, STORM_DIR + "/bin"],
         args=args,
-        childopts="-Dstorm.jar=" + jarfile)
+        jvmopts=["-Dstorm.jar=" + jarfile])
 
 def kill(*args):
     """Syntax: [storm kill topology-name [-w wait-time-secs]]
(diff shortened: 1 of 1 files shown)


bin/storm
```diff
@@ -208,12 +211,15 @@ def nimbus(klass="backtype.storm.daemon.nimbus"):
     (https://github.com/nathanmarz/storm/wiki/Setting-up-a-Storm-cluster)
     """
     cppaths = [STORM_DIR + "/log4j", STORM_DIR + "/conf"]
-    childopts = confvalue("nimbus.childopts", cppaths) + " -Dlogfile.name=nimbus.log -Dlog4j.configuration=storm.log.properties"
+    jvmopts = confvalue("nimbus.childopts", cppaths).split() + [
+        "-Dlogfile.name=nimbus.log",
```
nathanmarz, line 240: I don't think this is quite right, since the childopts could be something like -Dsome.config="aaa bbb".

bin/storm
```diff
@@ -225,12 +231,15 @@ def supervisor(klass="backtype.storm.daemon.supervisor"):
     (https://github.com/nathanmarz/storm/wiki/Setting-up-a-Storm-cluster)
     """
     cppaths = [STORM_DIR + "/log4j", STORM_DIR + "/conf"]
-    childopts = confvalue("supervisor.childopts", cppaths) + " -Dlogfile.name=supervisor.log -Dlog4j.configuration=storm.log.properties"
+    jvmopts = confvalue("supervisor.childopts", cppaths).split() + [
```
nathanmarz: Same as above. Should probably make this parsing piece its own function.

bin/storm
```diff
@@ -243,11 +252,14 @@ def ui():
     (https://github.com/nathanmarz/storm/wiki/Setting-up-a-Storm-cluster)
     """
     cppaths = [STORM_DIR + "/log4j", STORM_DIR + "/conf"]
-    childopts = confvalue("ui.childopts", cppaths) + " -Dlogfile.name=ui.log -Dlog4j.configuration=storm.log.properties"
+    jvmopts = confvalue("ui.childopts", cppaths).split() + [
+        "-Dlogfile.name=ui.log",
```
nathanmarz, line 281: Same as above.

nathanmarz:
Left a few comments on the pull request. Also, can you send me a contributor's agreement? Instructions at the bottom of this page: https://github.com/nathanmarz/storm/wiki/Contributing-to-Storm

jshuping:
Thanks guys, I ran into the same issue and was thinking of a similar patch!  Just need to do something about the end of shell()  (~line 194) as this change to using execvp() will mean that last os.system() in shell() will no longer get called.

ept:
@nathanmarz No problem, I've added a proper arguments parser: 0749adbe72360e38d82f33392b339977e472b9de. Contributor agreement coming shortly.

@jshuping Nicely spotted, thanks. Fixed: 08224b2a8a27ea6b5799bbd339da64662eda1b2a

shrijeet:
@ept  I tried your patch on 0.7.2 and got http://pastie.org/4076250 , I printed type() of various entities being concatenated in exec_storm_class method and found args to be of type tuple.  following change fixes it : 

``` python
] + jvmopts + [klass] + list(args)
```

I was running "storm list" when I first encountered this. Python version is 2.6

shrijeet:
@ept excuse me if I am wrong, python is not my forte.

ept:
@shrijeet Thanks — I had tested running Nimbus, supervisors etc. but hadn't tested the client-side commands. Your fix is good, I've added a commit for it to this branch.

@nathanmarz I think it's all stable now. Could you give it another look?

dctrwatson:
We use runit and having this issue for now I applied this directly, but would be nice to have it in mainline

ept:
We've also been using this patched version successfully on Linux and Mac OS. My only worry about `execvp` is that it may not work on Windows. The Python docs say it does, but it would be good to test it. Does anyone have a Windows machine handy on which they can test it?

bin/storm
```diff
@@ -82,12 +84,37 @@ def print_remoteconfvalue(name):
...
+        "(?: [^"\\] | \\.)*" |
+        '(?: [^'\\] | \\.)*' |
+        \\.
+    )+)''', re.VERBOSE)
+    args = re_split.split(string)[1::2]
+    args = [re.compile(r'"((?:[^"\\]|\\.)*)"').sub('\\1', x) for x in args]
+    args = [re.compile(r"'((?:[^'\\]|\\.)*)'").sub('\\1', x) for x in args]
+    return [re.compile(r'\\(.)').sub('\\1', x) for x in args]
+
+def exec_storm_class(klass, jvmtype="-server", jvmopts=[], extrajars=[], args=[], fork=False):
```
nathanmarz, line 106: We need to add fork=True for all the Storm commands, right? Maybe it would be better for fork to default to True, since the number of commands is likely to grow faster than the number of daemons.

ept, line 106: @nathanmarz  `fork=True` is only needed if the script needs to do something after the call to `exec_storm_class`. Currently, `shell` is the only command for which this is the case (it needs to remove the temporary jar file). For the other client commands, either way should be fine.

nathanmarz, line 106: Ah, I understand now. Thanks for clarifying.

nathanmarz:
Merged, thanks!

ept:
Awesome, thanks Nathan!

PR #239, opened 2012-05-31
storm pluggable scheduler

xumingming:
for issue #164


src/clj/backtype/storm/daemon/nimbus.clj
```diff
@@ -414,50 +386,117 @@
...
-      )
-    (merge stay-assignment reassignment)
-    ))
+;; in the future could allocate executors intelligently (so that "close" tasks reside on same machine)
+
+;; TODO: slots that have dead executor should be reused as long as supervisor is active
 
+;; public so it can be mocked out
+(defn compute-new-topology->executor->node+port [nimbus existing-assignments topologies scratch-topology-id]
+  (let [conf (:conf nimbus)
```
nathanmarz, line 498: This function has a lot going on in it, it would be good to clean this up to make it easier to read/maintain this code later on. It seems that a lot of these (into {} (for ...)) blocks can be refactored into their own functions.

src/clj/backtype/storm/daemon/nimbus.clj
```diff
@@ -414,50 +386,117 @@
...
+                                         ;; hide the dead-ports from the all-ports
+                                         ;; these dead-ports can be reused in next round of assignments
+                                         all-ports (-> sid
+                                                       all-slots
+                                                       (set/difference dead-ports)
+                                                       ((fn [ports] (map int ports))))
+                                         supervisor-details (SupervisorDetails. sid hostname scheduler-meta all-ports)]]
+                               {sid supervisor-details}))
+        cluster (Cluster. supervisors existing-scheduler-assignments)
+        scheduler (if (conf STORM-SCHEDULER)
```
nathanmarz: Should just create the scheduler once instead of recreating it each time you want to call schedule. This lets scheduler implementations keep state if they wish (I implemented this here on top of your pull request: https://github.com/nathanmarz/storm/commit/fce57833e3bb6e311878f160b283008808a607ed )

src/jvm/backtype/storm/scheduler/Cluster.java
```diff
@@ -0,0 +1,327 @@
...
+        slots.addAll(assignment.getExecutorToSlots().values());
+
+        return slots.size();
+    }
+
+    /**
+     * Assign the slot to the executors for this topology.
+     */
+    public void assign(WorkerSlot slot, String topologyId, Collection<ExecutorDetails> executors) {
+        SchedulerAssignment assignment = this.getAssignmentById(topologyId);
```
nathanmarz: This function should throw an exception if the slot is already assigned.

src/clj/backtype/storm/daemon/nimbus.clj
```diff
@@ -687,8 +734,8 @@
                         0
                         (conf NIMBUS-MONITOR-FREQ-SECS)
                         (fn []
-                          (doseq [storm-id (.active-storms (:storm-cluster-state nimbus))]
-                            (transition! nimbus storm-id :monitor))
+                          (when (conf NIMBUS-REASSIGN)
```
nathanmarz, line 785: There's actually a race condition on this line. Whenever invoking the scheduler you have to isolate the code within the submit-lock of Nimbus. I fixed this here: https://github.com/nathanmarz/storm/commit/488734aba329f7819c6f1fab218fdaea034302b9

nathanmarz:
Overall, this is really great work. I created a "scheduler" branch on top of this pull request with a couple changes (noted in the inline comments to this pull request). 

As for needed changes, I left notes within the inline comments. Additionally, I have two more comments:
1. I think the DefaultScheduler/EvenScheduler implementation could be better. Rather than pass in the keep-slots-fn, I think it's cleaner for the DefaultScheduler to just free the necessary slots, and then pass off the cluster/topologies to EvenScheduler which always just takes the remaining unassigned slots and evenly assigns them across nodes/workers.
2. Since this is a substantial new feature, there should be unit tests for it. See nimbus_test.clj for examples of scheduler tests.

nathanmarz:
Thanks for the changes. One more thing and then I'm ready to merge this in.

The only methods for manipulating slots should be "Cluster#assign" and "Cluster#freeSlot". Any other helper methods should be build on top of those methods. Currently there are these methods in the API which don't respect this:
1. Cluster#setAssignmentById
2. SchedulerAssignment#assign
3. SchedulerAssignmnent#unassignBySlot

There were some other unused methods that also violated this constraint but I removed them in the scheduler branch. The reason that this constraint is important is because it ensures that validations (like checking that the slot is not currently assigned when assigned to) are always applied.

I would recommend doing the following:
1. Get rid of the Cluster#setAssignmentById method and change the DefaultScheduler / EvenScheduler to call freeSlot and assignSlot directly
2. Split the SchedulerAssignment class into two classes: SchedulerAssignment and SchedulerAssignmentImpl. SchedulerAssignment will have the public read-only API, and SchedulerAssignmentImpl will be a subclass that exposes the mutable methods to be used internally by Cluster. Cluster will create SchedulerAssignmentImpl internally but only pass back SchedulerAssignment types in the getAssignmentById API.

nathanmarz:
@xumingming Also, I would recommend merging in the scheduler branch since I had to fix a minor conflict.

nathanmarz:
Ok, this is now merged in! Amazing contribution!

Just one note – there was one more thing I had to fix, which was the new feature that rebalance can change the number of workers/executors on the fly. I didn't have a test for this which was why you didn't see it. I added a test for this and then fixed the scheduler branch to pass the test: https://github.com/nathanmarz/storm/commit/5df7182d52d00734f665f8e9e14a6d6fac4e2b6f

PR #266, opened 2012-07-14
Issue72

xumingming:
implemented the java version of 'testing'


nathanmarz:
A few comments:
1. Let's make this work with LocalCluster. In the Java interface, you can make it take in a ILocalCluster. You can then add extra methods to LocalCluster not exposed in ILocalCluster that your code can use to extract the cluster map within it.
2. Let's write the TestingTest in Clojure, not Java :) That way it runs with the regular Storm test suite.
3. There's no need to expose the clusterState object in LocalCluster.

xumingming:
By "make this work with LocalCluster", you mean let the method defined in testing4j take a LocalCluster as param? to replace the `backtype.storm.testing.Cluster`?

xumingming:
My intention to write the unit test in java was it might also be a good sample to storm user telling them how to use the api, I will change them into clojure and move the sample code maybe into `storm-lib`

nathanmarz:
Yes, that's right. Replace the usage of backtype.storm.testing.Cluster with LocalCluster in the API.

Also, makes sense regarding converting the test to clojure for unit testing and moving the Java version to an example project.

xumingming:
converted the unit test to clojure and replace the `testing.Cluster` with `ILocalCluster`

src/jvm/backtype/storm/utils/Utils.java
```diff
@@ -342,4 +342,13 @@ public static CuratorFramework newCuratorStarted(Map conf, List<String> servers,
         buffer.get(ret, 0, ret.length);
         return ret;
     }
+    
+    public static List<Object> list(Object... objects) {
```
nathanmarz: This function is unnecessary. Just use the Values class which already does this. "new Values(obj1, obj2, obj3)". Values implements List.

src/jvm/backtype/storm/ILocalCluster.java
```diff
@@ -24,4 +24,6 @@
     StormTopology getTopology(String id);
     ClusterSummary getClusterInfo();
     TopologyInfo getTopologyInfo(String id);
+    Map getState();
+    Object getNimbus();
```
nathanmarz: I don't think this method is necessary on the interface. The only place that uses it is the Testing#submitLocalTopology method, and that can just be gotten rid of.

src/clj/backtype/storm/testing4j.clj
```diff
@@ -0,0 +1,130 @@
...
+  ([^Map result ^String componentId ^String streamId]
+     (let [stream-result (.get result componentId)
+           ret (if stream-result
+                 (find-tuples stream-result streamId)
+                 [])]
+       ret))
+  ([^Map result ^String componentId]
+     (-readTuples result componentId Utils/DEFAULT_STREAM_ID)))
+
+(defn -submitLocalTopology [^ILocalCluster cluster ^String topologyName ^Config config ^StormTopology topology]
```
nathanmarz: This method isn't needed, as users can just use LocalCluster#submitTopology method

xumingming:
Yeah, indeed `submitLocalTopology`, `Utils.list` are not needed, I removed them.

nathanmarz:
Merged! I made a few minor changes with respect to naming/types. I also uncommented some unit test assertions that were commented for some reason.

PR #270, opened 2012-07-22
More detailed error messages when the shell subprocess pipe closes.

vinodc:
I've added any ouput from the error stream of the subprocess to the detail message of the RuntimeException thrown to make it easier to figure out why the subprocess exited unexpectedly.

### DIFF
--- a/src/jvm/backtype/storm/utils/ShellProcess.java
+++ b/src/jvm/backtype/storm/utils/ShellProcess.java
@@ -2,18 +2,22 @@ package backtype.storm.utils;
 
 import backtype.storm.task.TopologyContext;
 import java.io.BufferedReader;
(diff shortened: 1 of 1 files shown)


nathanmarz:
Great idea. Can you send me a contributor agreement so I can merge this in? Instructions here (should only take a few minutes): https://github.com/nathanmarz/storm/wiki/Contributing-to-Storm

vinodc:
Sure. Sent you the agreement.

nathanmarz:
Thanks. This patch is going to make debugging multilang issues infinitely easier. I also made it so the error stream is captured and printed if the subprocess fails to launch.

vinodc:
Oh good idea. Yeah, it's been helpful during debugging.

PR #277, opened 2012-07-27
Update Curator Version

ptgoetz:
This is a simple bump to the latest version of netflix/curator that maintains compatibility with zookeeper 3.3.x.

**Notes related to upgrading to curator 1.1.x/zookeeper 3.4.x:**
- zookeeper 3.4.x uses Netty internally rather than NIO directly - this will impact the method for creating an in-process zookeeper in "zookeeper.clj" (mk-inprocess-zookeeper).
- The above issue can possibly be addressed by using some of the curator-test classes for creating in-proc zookeeper servers.
- Looks like there are some authentication options in the latest version of curator/zookeeper.

I tried a bump to curator 1.1.16/zookeeper 3.4.5 by reusing curator test classes (modified the the mk-inprocess-zookeeper method). It made the tests run excruciatingly slow with a bunch of stacktraces. Tests passed, but I didn't trust the result. I'll continue investigating at some point.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -12,7 +12,7 @@
                  [storm/libthrift7 "0.7.0"]
                  [clj-time "0.4.1"]
                  [log4j/log4j "1.2.16"]
-                 [com.netflix.curator/curator-framework "1.0.1"]
+                 [com.netflix.curator/curator-framework "1.0.16"]
                  [backtype/jzmq "2.1.0"]
                  [com.googlecode.json-simple/json-simple "1.1"]
                  [compojure "0.6.4"]


boneill42:
+1, need this to sort out our classpath issues.

ivolo:
+1

nathanmarz:
I can't accept pull requests that are just arbitrary upgrades of Storm's dependencies. These kinds of changes can have subtle, cascading effects. Last time I tried updating Curator it made the Storm unit tests take 10x longer, making development impossible. I will be updating Guava to v13 in the next release, but that's about it. If you guys want to use your own versions, your best bet is to implement #115.

ptgoetz:
I completely understand and agree.

My intent with the original pull request was to communicate some additional information regarding future issues that might impact storm with an upgrade to zookeeper 3.4.x, as well as the fact that an upgrade to the latest curator (zookeeper 3.3.x) is potentially a safe move for the storm 0.8.1 release.

For what it's worth, I'm running a soak test on a modest cluster with this patch (storm-0.8.1-wip2-SNAPSHOT) and we're seeing huge improvements. On storm 0.7.x we were averaging 3.5k writes to cassandra per second. With 0.8.x we're seeing 20k per second. So far we've processed several billion tuples, with 0 failures. 

That's a first for us. In the past we had to throttle back with topology_max_pending.

The ideal solution is to solve the elusive #115. But it's obviously not easy.

nathanmarz:
@ptgoetz Just to clarify, in your soak test, is this patch orthogonal to the performance improvements you're seeing? That is, is this patch just needed for library compatibility?

ptgoetz:
Library compatibility, for the most part.

I ran some regressions, and found that our performance improvements seemed to be based on the Storm 0.8.x migration, rather than the switch to astyanax.

Performance with astyanax vs. hector was roughly the same.

That being said, we're moving foreword with Astayanax, and Curator.

IMO, what's really needed is a good solution to #115, and as I mentioned, it is and will be elusive.

jasonjckn:
I'm merging all 3.4.5 zk related conversations to this issue https://github.com/nathanmarz/storm/issues/480

PR #288, opened 2012-08-11
Reverting the generics tightening because nested generics in java are limited

sjoerdmulder:
Sorry, reverting partially my change.

I'm currently busy with implementing a MongoDB State and It's just impossible to have something like:

``` java
state = new MongoState<TransactionalValue<S>>(template, TransactionalValue<S>.class);
cachedMap = new CachedMap<TransactionalValue<S>>(state, opts.localCacheSize);
return TransactionalMap.build(cachedMap);
```

And i need to pass the entityClass as argument because i need it for the mongo template to automaticly serialize / deserialize the pojo. 

``` java
state = new MongoState<TransactionalValue>(template, TransactionalValue.class);
cachedMap = new CachedMap<TransactionalValue>(state, opts.localCacheSize);
return TransactionalMap.<S>build(cachedMap);
```

it will be then

### DIFF
--- a/src/jvm/storm/trident/state/map/OpaqueMap.java
+++ b/src/jvm/storm/trident/state/map/OpaqueMap.java
@@ -8,24 +8,24 @@ import java.util.List;
 
 
 public class OpaqueMap<T> implements MapState<T> {
-    public static <T> MapState<T> build(IBackingMap<OpaqueValue<T>> backing) {
+    public static <T> MapState<T> build(IBackingMap<OpaqueValue> backing) {
         return new CachedBatchReadsMap<T>(new OpaqueMap<T>(backing));
     }
     
-    IBackingMap<OpaqueValue<T>> _backing;
+    IBackingMap<OpaqueValue> _backing;
     Long _currTx;
     
-    protected OpaqueMap(IBackingMap<OpaqueValue<T>> backing) {
+    protected OpaqueMap(IBackingMap<OpaqueValue> backing) {
         _backing = backing;
     }
     
     @Override
     public List<T> multiGet(List<List<Object>> keys) {
-        List<OpaqueValue<T>> curr = _backing.multiGet(keys);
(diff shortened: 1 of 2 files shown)


sjoerdmulder:
FYI i published the Mongo State: https://github.com/sjoerdmulder/trident-mongodb

nathanmarz:
Sorry, I don't understand the issue you're facing that necessitates this pull request. Can you explain further?

sjoerdmulder:
Will try :-) I assume that the usage of Transactional, Opaque and Non-transactional State should be transparent, so the wrapping `State<T>` has no real knowledge of what T is. (like your MemcachedState or MongoState i tried to build).

So here comes the problem i was facing in order for the Mongo document to be converted to a POJO class you would need to somehow pass the `Class<T> entity` to the constructor of the wrapping `State<T>` but because of java generics it must match the `<T>` you have for the class. So `new MongoState<TransactionalValue<T>>()` to be passed into the `TransactionalMap.build`  should have `TransactionalValue<T>.class` as argument, but that doesn't compile. Only `TransactionalValue.class`.

So i am a bit stuck there. Only way i could think of in fixing was reducing the generics in the build method, or maybe you have an idea?

sjoerdmulder:
Hi Nathan,

I have been doing some re-thinking what would be the best solution for this,  and actually i think i was wrong. The "framework" should provide the best way possible to allow passing generic in. 

Example: if a method accepts `Map<List<T> map` then you can also pass it `Map<List>` if you want to, it will generate a warning but you can ignore that.

But if a method accepts only `Map<List>` then you cannot pass in `Map<List<Object>>` you would have to cast it. 

So instead of having this generic loosening the method who calls the builders should loosenup their code (e.g. my MongoState in this case).

Could you undo the merge?

nathanmarz:
TBH, I prefer it this way. I don't like overusing generics so much, plus it doesn't really matter for these classes since users don't use them directly.

PR #293, opened 2012-08-15
switch log4j to logback

jasonjckn:


src/clj/backtype/storm/bootstrap.clj
```diff
@@ -28,7 +28,8 @@
                       [event :as event] [process-simulator :as psim]]))
      (require (quote [clojure.set :as set]))
      (require (quote [backtype.storm [stats :as stats] [disruptor :as disruptor]]))
-     (import (quote [org.apache.log4j PropertyConfigurator Logger]))
+     (import (quote [org.apache.log4j PropertyConfigurator]))
```
nathanmarz: should this be deleted?

logback/production.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<configuration scan="true" scanPeriod="60 seconds">
+ <appender name="A1" class="ch.qos.logback.core.rolling.RollingFileAppender">
+    <rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
+      <!-- daily rollover -->
+      <fileNamePattern>${storm.home}/logs/${logfile.name}.%d{yyyy-MM-dd}.log</fileNamePattern>
+
+      <!-- keep 2 days' worth of history -->
+      <maxHistory>2</maxHistory>
```
nathanmarz: Let's make the default 7 days.

src/clj/backtype/storm/bootstrap.clj
jasonjckn: This diff is correct. Should this delta be deleted? I don't understand. Importing org.apache.log4j.Logger should not be done. Are you asking about PropertyConfigurator?

nathanmarz:
One more thing – bin/to_maven.sh needs to be updated. It produces two artifacts for Maven, one that has the default logging configuration embedded (to console), and one that doesn't so that Storm can be used as a library (storm-lib).

nathanmarz:
Actually, you should do a find-all on the project for log4j. There's more that needs to be changed in bin/storm, and we should rename the redirect-stdio-to-log4j! function

nathanmarz:
Merged into 0.9.0 branch

PR #295, opened 2012-08-15
Update slf4j-* to 1.6.6 version

amuraru:
Hi Nathan,

Here's a quick fix to upgrade slf4j-\* deps for storm to the latest 1.6.6 version.
Namely, the most-wanted feature in slf4j-api 1.6.0 not available in1.5.\* is
http://slf4j.org/faq.html#paramException

slf4j-api is already there from thrift7 deps so I had to add it explicitly in the list of deps (backward compatible)

thanks
-adrian


nathanmarz:
The pull request that reworks the logging system to use logback upgrades slf4j to 1.6.5, so this will be included once that's merged in.

PR #300, opened 2012-08-16
log4j switched to logback & upgrade to leiningen 2

jasonjckn:


jasonjckn:
Actually, i'll update the old change list. No need to create new one.

PR #312, opened 2012-08-26
original conf file #243.

tjun:
for #243.
Added -conffile option to use original yaml file.

> storm jar allmycode.jar -conffile org.yaml org.me.MyTopology arg1 arg2 arg3
> ( Conf filename must be written in relative path from ~/.storm/)

Also added -confdir option to use orignal confdir, instead of ~/.storm .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
 ## Unreleased
 
- * Added high water mark to ZeroMQ sockets (defaults to 10000) configurable with zmq.hwm
  * Storm UI now uses nimbus.host to find Nimbus rather than always using localhost (thanks Frostman)
  * Added report-error! to Clojure DSL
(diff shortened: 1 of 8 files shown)


nathanmarz:
Sorry for the really late response to this. Can you send me a contributor agreement so that we can iterate on this? Instructions here: https://github.com/nathanmarz/storm/wiki/Contributing-to-Storm

tjun:
Thank you for your response. I think I had sent you the document. The mail was sent on Feb.9.
Should I  resend it?

nathanmarz:
You're right! I found it, and it looks good.

I'm not a fan of this implementation. I think we can make things substantially simpler.

I don't think there's any reason to make the conf dir configurable – we should just make the conf file configurable via the --config option. Here's what I propose:
1. In parse_config_opts, set a global variable called "CONFFILE" to the value of that option.
2. In exec_storm_class, set -Dstorm.conf.file option to the value of that global var.
3. In backtype.storm.utils.Utils#readStormConfig look for that environment variable. If it's set, use that filename. If it's not set, use the string "storm.yaml".

tjun:
Sorry for a lot of commits. 
I fixed code, so pick the last commit: 5cf1d90 please.

nathanmarz:
Thanks! I merged in that commit.

PR #321, opened 2012-09-05
Default to ipv4

gjcourt:
Use JVM_OPTS environment variable
Prefer ipv4 by default
Remove unused import

### DIFF
--- a/bin/storm
+++ b/bin/storm
@@ -4,7 +4,6 @@ import os
 import sys
 import random
 import subprocess as sub
-import getopt
 import re
 
 def identity(x):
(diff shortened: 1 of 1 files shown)


bin/storm
```diff
@@ -26,6 +25,9 @@ CONF_DIR = os.path.expanduser("~/.storm")
 STORM_DIR = "/".join(os.path.realpath( __file__ ).split("/")[:-2])
 CONFIG_OPTS = []
 
+def get_jvm_opts():
```
nathanmarz: What's this? This seems to be independent functionality from the purpose of this pull request, which is to default to ipv4?

gjcourt: It is independent. I left this in since allowing for overrides via JVM_OPTS seems to be the more standard java approach for application specific configuration. Since storm currently isn't ipv6 compatible, I think that the ipv4 fix made better sense as a part of the exec_storm_class configuration.

This piece probably makes sense as part of a larger refactor of this script, would you like it removed?

nathanmarz:
Yes, that piece should be done as a separate pull request so we can discuss that functionality separately.

nathanmarz:
Closing, since we're revamping the network layer and that will hopefully work for ipv6.

PR #327, opened 2012-09-10
Make it possible to submit topology in inactive state

jasonjckn:
https://github.com/nathanmarz/storm/issues/267


src/clj/backtype/storm/daemon/nimbus.clj
```diff
@@ -867,215 +867,226 @@
...
-          (when-not is
-            (throw (RuntimeException.
-                    "Could not find input stream for that id")))
-          (let [ret (.read is)]
-            (.put downloaders id is)
-            (when (empty? ret)
-              (.remove downloaders id))
-            (ByteBuffer/wrap ret)
-            )))
+    (letfn [(submit-topology-with-opts
```
nathanmarz: Why is it necessary to have another helper function here? Can't you have submitTopology just call submitTopologyWithOpts directly? Also, I think it's better for helper functions to just be their own functions rather than in the same form of the service handler via letfn.

test/clj/backtype/storm/integration_test.clj
```diff
@@ -276,6 +276,50 @@
...
+      (reset! bolt-prepared? false)
+      (reset! spout-opened? false)
+      
+      (submit-local-topology-with-opts (:nimbus cluster)
+        "test"
+        {}
+        (:topology tracked)
+        (SubmitOptions. TopologyInitialStatus/INACTIVE))
+      (.feed feeder [1])
+      (Thread/sleep 5000)
```
nathanmarz: Why is this necessary?

nathanmarz:
Looks really good, except for a couple minor comments.

test/clj/backtype/storm/integration_test.clj
jasonjckn: why is what necessary?

src/clj/backtype/storm/daemon/executor.clj
```diff
@@ -343,19 +344,19 @@
     ret
     ))
 
+(defn sleep-until-true [wait-fn]
```
nathanmarz: Isn't it simpler to just use "while" here?

jasonjckn: I learned something new in clojure core.  that hasn't happened for a while :)

PR #338, opened 2012-09-21
add validation for the projection field

xumingming:
for issue #324


nathanmarz:
This isn't quite what we want (I was wrong about doing this in the projection factory). The problem is that the projection factory  isn't created until the workers launch, but we want these errors to happen on the line the mistake is made (during topology construction). Additionally, we want errors to be thrown for invalid groupBy or partitionBy calls during topology construction.

xumingming:
Ok, understand your concern now.

PR #339, opened 2012-09-21
add api for specifing name for stream/bolt

xumingming:
for issue #332


nathanmarz:
I decided to go with a different implementation with somewhat different semantics:
1. Names for internal nodes are appended so you know everything running in that bolt
2. Setting a name carries through to future nodes as well

Here's the commit: https://github.com/nathanmarz/storm/commit/60dc00149b1eb663437b8355c006fc4ccf43a6b4

PR #344, opened 2012-09-26
log topology submission issues

jasonjckn:
A storm user was not printing exceptions thrown by StormSubmitter, wasted a bit of time because nimbus.log had no info about it.


nathanmarz:
This needs to rethrow the exception so it gets to the client. Also, can you do a quick test to make sure the exception  reaches the client as is and doesn't get wrapped in a RuntimeException, as Clojure does in some situations?

PR #347, opened 2012-09-30
Zk update

Frostman:
Hi, 

Curator 1.0.x is end of life now and no longer maintained. So, I try to update it to the latest version (1.0.1 -> 1.1.18). 

All tests (lein test) are passed successfully (at OS X 10.8 and Ubuntu 12.04.1).

Then I updated storm to the version with new curator at my dev cluster with Zk 3.3.6 and storm worked good after it. Topologies are still working and actions like submit/kill/rebalance topology work good.

After it I updated Zk to 3.4.4 (latest stable) version  at the same cluster and storm is working good after it and all actions are still working.

So, I think that we can update Curator (and Zk) harmless to the latest versions.

P.S. The are a lot of issues fixed in Curator 1.1.x and in Zk 3.4.x branches.


Frostman:
Are there any obvious disadvantages of Curator upgrade?

Frostman:
lein test passed successfully, build: http://f.frostman.ru/tsc-dists/storm-0.8.2-wip11-zk-update.zip

nathanmarz:
Last time we tried upgrading Curator it made the unit tests about 10x slower. So we're being fairly cautious about upgrading it. Additionally, as it's so core to how Storm works, it requires a lot of testing in a production-like environment to be satisfied that it meets Storm's robustness requirements. One thing that would be helpful is explaining what the upgrade actually buys Storm. What are the improvements and fixes that are relevant?

jasonjckn:
I'm merging all 3.4.5 zk related conversations to this issue https://github.com/nathanmarz/storm/issues/480

PR #348, opened 2012-09-30
Topology base actions buttons added to the UI

Frostman:
Some screenshots: http://f.frostman.ru/storm/ui-tplg-actions/


nathanmarz:
What happens if someone doesn't have Javascript enabled on the browser? Is it then impossible to execute the commands (this is desired behavior – these commands should be impossible without confirmation)?

Frostman:
All action buttons are implemented using JavaScript, so, if someone will use browser with disabled JS and click the action button then nothing happens. 

Additionally, "js-only" css class has been implemented for elements that should be displayed only if JavaScript is enabled, so, all buttons (including Show/Hide System Stats) and "Topology actions" section will be hidden for browsers with disabled JS.

builded storm: http://f.frostman.ru/tsc-dists/storm-0.8.2-wip5-ui-tplg-actions.zip

nathanmarz:
Hmm, this no longer merges cleanly. Can you merge in master and fix any conflicts?

Frostman:
Ok, will do it.

Frostman:
Done

Frostman:
p.s. lein test passed successfully, new build: http://f.frostman.ru/tsc-dists/storm-0.8.2-wip10-ui-tplg-actions.zip

nathanmarz:
Looks great. Let's add one more thing to it though. This would be much more useful if you could override the wait time for KILL and REBALANCE. The most common case is just doing -w 0, so how about adding IMMEDIATE KILL and IMMEDIATE REBALANCE commands? Or you could add some sort of interface for specifying the wait time.

Frostman:
I think that confirm can be replaced with prompt for rebalance and kill actions. Here is an example of how it can looks like: http://jsfiddle.net/3AdrL/ . What do you think about it? 

P.S. Default value will be taken from conf.

nathanmarz:
That sounds good to me. The default value should be the topology message timeout.

Frostman:
Ok, I will implement it.

Frostman:
Done.

lein test passed successfully, new build: http://f.frostman.ru/tsc-dists/storm-0.8.2-wip10-ui-tplg-actions-2.zip

nathanmarz:
Merged! Thanks for the great contribution.

PR #361, opened 2012-10-11
Fix FileNotFoundException during cleanup

MichaelBlume:
See issue #356

Code already checks for existence of file, but sometimes we get
FileNotFoundException anyway. Honestly not sure why, but this fixes it.

### DIFF
--- a/src/clj/backtype/storm/util.clj
+++ b/src/clj/backtype/storm/util.clj
@@ -410,7 +410,13 @@
 (defn rmr [path]
   (log-debug "Rmr path " path)
   (when (exists-file? path)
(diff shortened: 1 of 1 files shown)


MichaelBlume:
Upon further investigation, this looks more like a band-aid than a fix -- see #356

nathanmarz:
I'm actually fine with wrapping rmr in a FileNotFound. @MichaelBlume , can you send me a contributor agreement so I can merge this in? Should only take a couple minutes, instructions are here: https://github.com/nathanmarz/storm/wiki/Contributing-to-Storm

PR #364, opened 2012-10-14
Make UI friendly to repl development by not joining jetty thread

gsilk:
This lets you run the jetty server alongside the repl, which makes it easier to do development on the UI. The use of 'defonce' lets you recompile without worrying that the server gets redefined. The following commands start/stop the server from the repl:

(.start server)
(.stop server)

### DIFF
--- a/src/clj/backtype/storm/ui/core.clj
+++ b/src/clj/backtype/storm/ui/core.clj
@@ -728,5 +728,7 @@
   (handler/site main-routes)
  )
 
-(defn -main []
-  (run-jetty app {:port (Integer. (*STORM-CONF* UI-PORT))}))
+(defn start-server! [] (run-jetty app {:port (Integer. (*STORM-CONF* UI-PORT))
+                                       :join? false}))
+
+(defn -main [] (start-server!))


nathanmarz:
Does this mean that a jetty server is created during compilation? I typically don't like putting anything beyond simple data in top level forms, preferring to make the launching just a function, e.g. `(defn launch-server! [...])`

Also, why is the change that sends the var for app rather than the value necessary? What does that do?

gsilk:
We definitely don't want jetty running compile-time, I'm not sure what I was thinking last night. Also, accessing the var was left over from an unrelated change, I just forgot to remove it before submitting :(

How about something along these lines:

``` clojure
(defn start-server! [] (run-jetty app {:port (Integer. (*STORM-CONF* UI-PORT))
                                       :join? false}))

(defn -main [] (start-server!))
```

nathanmarz:
That seems fine. So in that case Jetty just runs in another thread as a non-daemon thread?

gsilk:
Yep, and so you can continue to play in the repl with Jetty running.

nathanmarz:
Sounds good, can you update the pull request?

gsilk:
Done!

nathanmarz:
Merged, thanks

PR #366, opened 2012-10-16
Include a warning about Leiningen 2.x in docs

gsilk:
Currently "lein (uber)jar" silently fails when using lein >= 2, and "lein test" complains about missing files on the classpath.

This is a pain when you're trying to get your dev environment setup for Storm, after following every instruction to a tee, only to have nothing work. As long as Storm is incompatible with lein 2.x, a big warning should be included somewhere in the docs telling the user to use lein 1.x

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,3 +1,8 @@
+(def lein-version (System/getenv "LEIN_VERSION"))
+(if-not (re-find #"^1\..*$" lein-version)
+  (do (println (str "ERROR: requires Leiningen 1.x but you are using " lein-version))
+    (System/exit 1)))
+
 (defproject storm "0.8.2-wip11"
   :source-path "src/clj"
   :test-path "test/clj"
@@ -36,4 +41,3 @@
   :extra-classpath-dirs ["src/ui"]
   :aot :all
 )
-


nathanmarz:
You're referring to compiling Storm itself? The 0.9.0 branch is upgraded to lein2.

gsilk:
Still, it would be good to have a warning in place until 0.9.0 is merged into master. New contributors will do their work off of master, and currently, if they go and download leiningen and start trying to contribute, _nothing_ will work (and there won't be any sensible errors describing why). That's a great way to give all your newcomers a migraine. Just sayin'.

nathanmarz:
Fair enough. Is there a way to set a required lein version flag in the project.clj?

gsilk:
Dug around and found that Leiningen exports its version into the environment. Code attached.

gsilk:
I also looked at what was available in defproject -- unfortunately lein only supports a minimum version check.

nathanmarz:
Merged, thanks. BTW, in the future you should put your pull request in its own branch (not master) so that you can keep it isolated from any other changes you want to make.

PR #371, opened 2012-10-18
Nimbus storage

Frostman:
Nimbus storage abstraction implementation, this is required for Nimbus HA implementation (#360).

I'd run only "lein test" and some handmade sanity tests on my laptop.


Frostman:
build: http://f.frostman.ru/tsc-dists/storm-0.8.2-wip11-storage.zip

Frostman:
Any comments?

nathanmarz:
@Frostman Haven't had a chance to review this deeply – I'll try to get to that this week.

Frostman:
@nathanmarz Thank you!

Frostman:
Mmm, any thoughts on it?

src/jvm/backtype/storm/Config.java
```diff
@@ -199,8 +199,16 @@
...
-    
+
+    public static String NIMBUS_STORAGE = "nimbus.storage";
+
+    /**
+     * A directory on the local filesystem used by Nimbus default storage to
+     * store files. If this property is undefined then STORM_LOCAL_DIR will
+     * be used.
+     */
+    public static String NIMBUS_LOCAL_DIR = "nimbus.local.dir";
```
nathanmarz: Why is a separate config needed here? Just always use STORM_LOCAL_DIR in the case of default storage.

src/clj/backtype/storm/nimbus/storage.clj
```diff
@@ -0,0 +1,68 @@
+(ns backtype.storm.nimbus.storage
+  (:import [java.io InputStream OutputStream FileInputStream FileOutputStream File])
+  (:import [java.util List Map])
+  (:import [org.apache.commons.io FileUtils IOUtils])
+  (:use [backtype.storm log config util])
+  (:import [backtype.storm.utils Utils])
+  (:import [backtype.storm.nimbus INimbusStorage]))
+
+(defn create-local-storage [conf]
```
nathanmarz, line 9: I think there are problems below with regards to separators between the storm root and the path. You should join the paths together with a "/".

nathanmarz:
Sorry for the super long delay! Overall this looks good, there's just a couple things that need updating as I indicated in the line notes.

I was on the fence about moving the rooting logic into the storage abstraction itself, but I think it makes sense.

src/clj/backtype/storm/config.clj
```diff
@@ -74,12 +74,26 @@
...
+
 (defn master-stormdist-root
   ([conf]
      (str (master-local-dir conf) "/stormdist"))
   ([conf storm-id]
      (str (master-stormdist-root conf) "/" storm-id)))
 
+(defn nimbus-storage-stormdist-root
+  ([]
+    (str "/stormdist"))
```
nathanmarz: The "str" here seems unnecessary.

src/clj/backtype/storm/nimbus/storage.clj
Frostman, line 9: All relative paths starts with "/", so, I think that there are no problems with separators between root ant relative path.

All relative paths are generated using:

``` clj
(defn nimbus-storage-stormdist-root
  ([]
    "/stormdist")
  ([storm-id]
    (str (nimbus-storage-stormdist-root) "/" storm-id)))
```

src/jvm/backtype/storm/Config.java
Frostman: I think that it's good to have an ability to define separated place for default storage because storage api is the separated abstraction. 

Additionally this config used to pass nimbus storage path to supervisors in local mode. When we use storage API only relative paths are stored into ZooKeeper and, so, we can't get topology jars from supervisors in local mode without the knowledge where is nimbus temp directory located. 

If you don't like separated config, we can hide it from users and don't define it in the configs.

src/clj/backtype/storm/testing.clj
```diff
@@ -108,8 +109,8 @@
                            daemon-conf
                            {STORM-CLUSTER-MODE "local"
                             STORM-ZOOKEEPER-PORT zk-port
-                            STORM-ZOOKEEPER-SERVERS ["localhost"]})
-        nimbus-tmp (local-temp-path)
+                            STORM-ZOOKEEPER-SERVERS ["localhost"]
+                            NIMBUS-LOCAL-DIR nimbus-tmp})
```
Frostman: Here is the place where we pass nimbus temp directory location to all daemons (nimbus and supervisors) in local mode.

src/jvm/backtype/storm/Config.java
nathanmarz: I understand that point, but I'd rather avoid unnecessary configuration parameters. The path used for local storage implementation should always just be STORM_LOCAL_DIR + "/nimbus/storage" or something like that.

src/clj/backtype/storm/nimbus/storage.clj
nathanmarz, line 9: I'm more comfortable if we add code that checks that the first character is a slash, and adds one of it isn't.

nathanmarz:
Responded to your comments.

src/jvm/backtype/storm/Config.java
Frostman: Done, `nimbus.local.dir` is used only for local mode now. The path used for local storage impl is `STORM_LOCAL_DIR + "/nimbus/"`.

src/clj/backtype/storm/nimbus/storage.clj
Frostman, line 9: Done, `join-paths` function has been added and used instead of `str` to build absolute path.

Frostman:
It looks like all issues has been resolved. Any other comments?

Thank you.

Frostman:
Any notes?

Thank you.

nathanmarz:
Looks good. I merged it into the nimbus-ha branch. Let's do the rest of the pull requests to that branch to finish this feature.

PR #378, opened 2012-10-20
Use resources evenly when managing multiple topologies

rnfein:
https://github.com/nathanmarz/storm/issues/359

Updated `EvenScheduler/sort-slots` to take slots first from supervisors with the most available slots.


nathanmarz:
This isn't right. It needs to sort the slots based on how many slots are used on that host, not based on how many slots are available.

Also, your test case is too specific. You should just check that the second supervisor has two ports assigned, not the specific ports that are assigned. It would also be good to have a test case for a more complicated set up, like 4 workers on one machine, 2 on 2 more, and 1 on another, and a new topology of 3 workers is assigned. It should check that the new distribution of workers is 4, 3, 3, 3.

nathanmarz:
Closing since the issues were never fixed.

PR #380, opened 2012-10-23
Add a way to send an exception through shellbolt

Lazyshot:
This provides a valid command for shell processes send back an exception or general error message.  This would allow a viewable error in the UI for ease in production.

I continue to see "failed tuples" in my shell processes, but I can not see the underlying reason for them without seeking out the individual worker log and line.

This change should allow a user to see the passed exception from the shell process straight to the UI for easy diagnoses and debugging.

The only edit I would make is to either move this functionality to the "log" command or create a specific exception class like ShellTupleException for more clear messages.

### DIFF
--- a/src/jvm/backtype/storm/task/ShellBolt.java
+++ b/src/jvm/backtype/storm/task/ShellBolt.java
@@ -181,9 +181,7 @@ public class ShellBolt implements IBolt {
 
     private void handleError(Map action) {
         String msg = (String) action.get("msg");
-        if (msg != null) {
-            _collector.reportError(new Exception("Shell Process Exception: " + msg));
-        }
+        _collector.reportError(new Exception("Shell Process Exception: " + msg));
     }
 
     private void handleEmit(Map action) throws InterruptedException {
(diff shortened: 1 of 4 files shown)


src/jvm/backtype/storm/task/ShellBolt.java
```diff
@@ -177,6 +179,13 @@ private void handleFail(Map action) {
         _collector.fail(failed);
     }
 
+    private void handleError(Map action) {
+        String msg = (String) action.get("msg");
+        if (msg != null) {
```
nathanmarz: Why do you have to check for null here?

nathanmarz:
Looks good overall, except for one comment. Could you also update the storm.py and storm.rb adapters to automatically report errors when an exception bubbles out of the main loops?

Lazyshot:
I added reportError methods to each of the languages and changed each usage of log to reportError in the proper contexts.

nathanmarz:
@Lazyshot Actually, I need you to sign a Contributor Agreement. For some reason I thought you had signed one already. Instructions here: https://github.com/nathanmarz/storm/wiki/Contributing-to-Storm (should only take a few minutes). Otherwise I'm going to have to unmerge this.

Lazyshot:
I just sent over the Contributor Agreement.

nathanmarz:
Thanks, I got it. Looks good.

PR #387, opened 2012-10-27
add projection validation for stream methods

xumingming:
for issue #324


nathanmarz:
Merged, thanks! This is going to be very useful.

PR #394, opened 2012-11-02
kill --all flag

Frostman:
Simple way to kill all topologies in Storm cluster.

build: http://f.frostman.ru/tsc-dists/storm-0.8.2-wip15-killall.zip

### DIFF
--- a/src/clj/backtype/storm/command/kill_topology.clj
+++ b/src/clj/backtype/storm/command/kill_topology.clj
@@ -4,11 +4,24 @@
   (:import [backtype.storm.generated KillOptions])
   (:gen-class))
 
+(defn kill-topology [nimbus name opts]
+  (.killTopologyWithOpts nimbus name opts)
+  (log-message "Killed topology: " name))
+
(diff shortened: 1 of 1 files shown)


nathanmarz:
I don't think this is important to have as a core feature in Storm. If people want to do this they can write their own scripts for it.

Frostman:
I thought it was great to have additional facilities out of the box. I'm sorry to bother you.

nathanmarz:
No it's fine, but sometimes I have to reject features that I think aren't useful enough to be in the core project. I always welcome new ideas.

tas50:
I'd argue that this is a code change worth doing. We maintain a publicly available Chef recipe to deploy Storm at my company.  It would be very nice to be able to kill all running topologies as part of the deploy process.  Without this we can't fully automate the process of deploying storm.  We have added our own code to kill our specific topology, but this isn't something that can be contributed to the greater community.  I imagine that anyone trying to automate deploys of storm using either Puppet or Chef would run into a similar problem.

PR #395, opened 2012-11-08
catch-errors middleware added

Frostman:
Display stack trace instead of plain white screen when exception happens in action method of UI.
It's very useful for me to see exception in browser. For example, I can understand when nimbus isn't available or topology with specified name isn't exists, etc.

Screenshot: http://f.slukjanov.name/storm/ui-errors-demo.png

P.S. we can write tons of handlers to write error message like "Nimbus isn't available", but I think that this information is easy parseable from stack traces.


nathanmarz:
This is great, thanks.

PR #399, opened 2012-11-14
set supervisor_id as optional

jhnstn:
I found that by setting the supervior_id to optional in the struct definition I am able to get a response  from

``` ruby
client.getClusterInfo()
```

When it is set as required I am getting 

```
Thrift::Exception: Required field supervisors is unset!
```


nathanmarz:
That field should always be set, so this is not a fix. Does the Storm UI work for you?

PR #404, opened 2012-11-20
Proxy getCodeDir to the TopologyContext in TridentOperationContext

ebroder:


nathanmarz:
Can you send me a contributor agreement? I need it even for small changes. Info here (should only take a few minutes to fill out): https://github.com/nathanmarz/storm/wiki/Contributing-to-Storm

ebroder:
Sorry for the delay on this. I had a chat with our general counsel, and we're not totally comfortable with the CLA as it's currently written (the patent license is particularly concerning to us).

I actually ended up not needing this for now (switched from Trident to straight Storm), and I don't really want to get into complex negotiations over a 3-line patch, so I'm going to close this. I expect that as I use Storm more extensively, I'll run into more significant potential changes. We can figure out what makes sense then, but I don't think it makes sense to waste time now.

Thanks

PR #409, opened 2012-11-26
fixed missing generic type params for trident

codefromthecrypt:
fixed a series of generic type params in trident, typically where Object  or unspecified types were used when a value type is warranted.

You'll notice a switch from List<List<Object> to List<? extends List<Object>>.  This is to permit backtype.storm.tuple.Values without casting.


nathanmarz:
Don't think the extra typing is that valuable, so closing.

PR #416, opened 2012-12-07
Making the to_maven.sh script support development mode

mbautin:
Adding a --dev option to the bin/to_maven.sh script that allows to build and install a storm-lib jar into the local Maven repository.


nathanmarz:
Not interested in making these scripts more complex. lein install is totally sufficient for development.

PR #424, opened 2012-12-17
Drain error stream periodically to avoid dead lock between ShellBolt and shell process

lyogavin:
In this fix, I periodically drain the error stream so that if the buffer of error stream of shell process becomes full, it won't be blocked there thus result it dead lock between ShellBolt and the shell process.

I put the error stream reading logic in the writing thread as there's no non-blocking api to read error stream, if we read error stream in reading thread, we need to do some looping and sleeping, which impact the performance of normal reading. And it's not worthy to create another thread to do this. I think normally we can assume there won't be too much data output to error stream, so just periodically drain the error stream in writing thread seems enough.

### DIFF
--- a/src/jvm/backtype/storm/task/ShellBolt.java
+++ b/src/jvm/backtype/storm/task/ShellBolt.java
@@ -123,6 +123,8 @@ public class ShellBolt implements IBolt {
                         if (write != null) {
                             _process.writeMessage(write);
                         }
+                        // drain the error stream to avoid dead lock because of full error stream buffer
+                        _process.drainErrorStream();
                     } catch (InterruptedException e) {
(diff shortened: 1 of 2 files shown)


nathanmarz:
Before I can look at this, I'm going to need you to send me a contributor agreement. Instructions here (should only take a few minutes): https://github.com/nathanmarz/storm/wiki/Contributing-to-Storm

lyogavin:
Thanks, Nathan. I need to go through some internal process before signing that, after that I'll get back to you.

lyogavin:
Nathan, i've signed the CLA and mailed to you. Can you please see if you have received it? If so, can we move ahead?

PR #428, opened 2013-01-07
Fix issue #403

herberteuler:
Hi Nathan, I believe this patch fixes issue #403.

You are right, it was not race condition, but rather some already closed sockets might be left in the new cached sockets map, due to the wrong invocation to dissoc.


xiaokang:
Hi @herberteuler and @nathanmarz . I verified that cached sockets map does not remove closed zmq socket by logging. But I can not explain why the closed zmq socket will be used since the :cached-task->node+port map is updated correctlly.

herberteuler:
@xiaokang You may need to check if the old node+port that the dead worker used has been reassigned to run a new worker.  If this happens, the same old node+port will still be needed and used.

PR #431, opened 2013-01-07
remove timeout drpc request from its function's request queue.

xiaokang:
for issuse #430.

add warn log for timeout drpc request and remove it form request queue.


xiaokang:
Hi @nathanmarz , is the code OK to fix #430? It aims to prevent a timed-out drpc request wasting drpc server memory and spout/bolt computing resource since result for timed-out request will be discard by drpc server.

xiaokang:
just ping.

PR #443, opened 2013-01-22
build: clean possible warning when running build_release.sh

ccding:
Always use -f argument when using rm, to avoid "file not exist" warning in the
first time build.


nathanmarz:
Closing, as someone else seems to have made this change.

PR #453, opened 2013-02-02
Set thread name for spout and bolt threads

noslowerdna:
This patch simply appends the component id for spouts and bolts to the thread name. We have found that having descriptive thread names for spout and bolt worker threads is quite helpful when analyzing jstack output or using a profiler such as VisualVM.


nathanmarz:
Thanks, merged. I removed the test since it's not really necessary and tests shouldn't rely on time the way it did.

PR #454, opened 2013-02-02
Update zookeeper / curator etc

GregBowyer:
Noting that the previous attempt was reverted for performance reasons,
this is a new attempt at this. Hopefully we will be able to track down
the performance regression.

I am more that willing to work through performance gotchas here, so this doesnt have to be merged immediately.


nathanmarz:
Closing because we've already done this in a separate branch and will be merging that in soon.

PR #460, opened 2013-02-09
fixed build failure on master

vbajaria:
lein jar was throwing ClassNotFoundException for LocalCluster and LocalDRPC, fixed it by putting require at the top of the file.

I saw 2 other people on the list face the same issue and I had similar issues when running it locally or using storm-deploy


nathanmarz:
I'm not able to reproduce this. Please open an issue for this so we can determine the cause before patching.

PR #461, opened 2013-02-10
Plugin for tuple serialization

anfeng:
Nathan, enclosed please find the proposed change for tuple serialization as discussed at issue #393. Please review.

### DIFF
--- a/src/jvm/backtype/storm/serialization/SerializationFactory.java
+++ b/src/jvm/backtype/storm/serialization/SerializationFactory.java
@@ -34,23 +34,19 @@ public class SerializationFactory {
         Kryo k = kryoFactory.getKryo(conf);        
         k.register(byte[].class);
 
-        /* tuple payload serializer could be specified via configuration */
+        /* tuple payload serializer is specified via configuration */
         String payloadSerializerName = (String)conf.get(Config.TOPOLOGY_TUPLE_SERIALIZER);
-        if (payloadSerializerName==null) 
-            k.register(ListDelegate.class, new ListDelegateSerializer()); //use default payload serializer
-        else {
-            try {
-                Class serializerClass  = Class.forName(payloadSerializerName);
-                Serializer serializer = resolveSerializerInstance(k, ListDelegate.class, serializerClass, conf);
-                if (serializer == null)
-                    k.register(ListDelegate.class, new ListDelegateSerializer());
-                else 
-                    k.register(ListDelegate.class, serializer);
-            } catch (ClassNotFoundException ex ){
-                LOG.error(ex + " Could not load class in class path: " + payloadSerializerName);
+        try {
+            Class serializerClass  = Class.forName(payloadSerializerName);
+            Serializer serializer = resolveSerializerInstance(k, ListDelegate.class, serializerClass, conf);
+            if (serializer == null)
                 k.register(ListDelegate.class, new ListDelegateSerializer());
-            } 
-        }
+            else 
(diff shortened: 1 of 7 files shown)


src/jvm/backtype/storm/serialization/SerializationFactory.java
```diff
@@ -32,7 +33,25 @@ public static Kryo getKryo(Map conf) {
...
         k.register(byte[].class);
-        k.register(ListDelegate.class);
+
+        /* tuple payload serializer could be specified via configuration */
+        String payloadSerializerName = (String)conf.get(Config.TOPOLOGY_TUPLE_SERIALIZER);
+        if (payloadSerializerName==null) 
+            k.register(ListDelegate.class, new ListDelegateSerializer()); //use default payload serializer
+        else {
+            try {
+                Class serializerClass  = Class.forName(payloadSerializerName);
```
nathanmarz: If the class isn't found there should be a runtime error. The user has done something wrong in this case and should be notified – silently using the default serializer is confusing and could be dangerous (e.g., the user was expecting encryption).

src/jvm/backtype/storm/serialization/SerializationFactory.java
```diff
@@ -32,7 +33,25 @@ public static Kryo getKryo(Map conf) {
         IKryoFactory kryoFactory = (IKryoFactory) Utils.newInstance((String) conf.get(Config.TOPOLOGY_KRYO_FACTORY));
         Kryo k = kryoFactory.getKryo(conf);        
         k.register(byte[].class);
-        k.register(ListDelegate.class);
+
+        /* tuple payload serializer could be specified via configuration */
+        String payloadSerializerName = (String)conf.get(Config.TOPOLOGY_TUPLE_SERIALIZER);
```
nathanmarz, line 38: Instead of having a special case for the name being null, let's just add a line to defaults.yaml. Then the code will be simpler.

src/jvm/backtype/storm/serialization/SerializationFactory.java
```diff
@@ -32,7 +33,25 @@ public static Kryo getKryo(Map conf) {
         IKryoFactory kryoFactory = (IKryoFactory) Utils.newInstance((String) conf.get(Config.TOPOLOGY_KRYO_FACTORY));
         Kryo k = kryoFactory.getKryo(conf);        
         k.register(byte[].class);
-        k.register(ListDelegate.class);
```
nathanmarz, line 35: Why get rid of this registration?

nathanmarz:
Mostly looks good – made a few comments inline.

src/jvm/backtype/storm/serialization/SerializationFactory.java
anfeng, line 35: This statement is needed if we use default serializer (http://kryo.googlecode.com/svn/api/v2/index.html). In our case, we will use seriaizer class specified via configuration, and thus we should remove this statement.

src/jvm/backtype/storm/serialization/SerializationFactory.java
anfeng, line 38: agree

src/jvm/backtype/storm/serialization/SerializationFactory.java
anfeng: agree

anfeng:
Please take a look at the revision. I modified defaults.xml and serialization/SerializationFactory.java.

PR #465, opened 2013-02-12
Initialising ReducerAggregator if (when) state is null.

lorcan:
I've updated `ReducerValueUpdater` to initialise its `ReducerAggregator` whenever `stored` is null. This solution assumes that `stored` will only be null the first time `update` is run. Attempts to address nathanmarz/storm#464. (Possibly this solution is a little too dumb/naive?)
[This  problem was discussed on the mailing list](https://groups.google.com/d/topic/storm-user/3kQU_8FO1xg/discussion)

### DIFF
--- a/src/jvm/storm/trident/state/ReducerValueUpdater.java
+++ b/src/jvm/storm/trident/state/ReducerValueUpdater.java
@@ -15,7 +15,7 @@ public class ReducerValueUpdater implements ValueUpdater<Object> {
 
     @Override
     public Object update(Object stored) {
-        Object ret = stored;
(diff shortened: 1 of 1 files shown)


nathanmarz:
Hi Lorcan, can you please send me a filled out CLA? I can't accept contributions without it. CLA's provide a legal basis for accepting contribution and protect the project and community. The CLA is here:

https://dl.dropbox.com/u/133901206/storm-apache-style-cla.txt

The easiest way to send it to me is to use HelloFax as you won't need to print/scan anything (should only take a few minutes).

PR #469, opened 2013-02-13
Apply Thrift SASL client/server framework for authentication/authorization/audit

anfeng:
For issue #397, enclosed please find the basic authentication/authorization/audit framework based on Thrift SASL implementation.  Nimbus integration and DRPC integration will be pushed as follow-up pull requests.

Authentication:
    - Support SASL authentication framework
    - Implemented SASL mechanisms: Anonymous, Kerberos, Digest
    - JAAS configuration

Authorization:
    - Enable plugins for Nimbus authorization against topology submission/killing/activation/deactivation/rebalance.
    - Plugin will have access to request context such as remote IP address, user name/principal etc.

Audit:
    - Simple log of operations performed by principals

### DIFF
--- a/src/jvm/backtype/storm/security/auth/AuthUtils.java
+++ b/src/jvm/backtype/storm/security/auth/AuthUtils.java
@@ -1,58 +1,47 @@
 package backtype.storm.security.auth;
 
+import backtype.storm.Config;
 import javax.security.auth.login.Configuration;
 import javax.security.auth.login.AppConfigurationEntry;
+import java.security.NoSuchAlgorithmException;
+import java.security.URIParameter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import backtype.storm.Config;
+import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.net.URI;
 import java.util.Map;
 
 public class AuthUtils {
+    private static final Logger LOG = LoggerFactory.getLogger(AuthUtils.class);
     public static final String LOGIN_CONTEXT_SERVER = "StormServer"; 
     public static final String LOGIN_CONTEXT_CLIENT = "StormClient"; 
     public static final String SERVICE = "storm_thrift_server";
 
     /**
-     * Construct a JAAS configuration object per storm configuration file
+     * Construct a JAAS configuration object per storm configuration file 
      * @param storm_conf Storm configuration 
-     * @return
+     * @return JAAS configuration object
      */
-    public static synchronized Configuration GetConfiguration(Map storm_conf) {
-        //retrieve system property 
-        String orig_loginConfigurationFile = System.getProperty("java.security.auth.login.config");
-
-        //try to find login file from Storm configuration  
-        String loginConfigurationFile = (String)storm_conf.get("java.security.auth.login.config");
-        if ((loginConfigurationFile==null) || (loginConfigurationFile.length()==0))
-            loginConfigurationFile = orig_loginConfigurationFile;
-
+    public static Configuration GetConfiguration(Map storm_conf) {
         Configuration login_conf = null;
+
+        //find login file configuration from Storm configuration  
+        String loginConfigurationFile = (String)storm_conf.get("java.security.auth.login.config");
         if ((loginConfigurationFile != null) && (loginConfigurationFile.length()>0)) { 
-            //We don't allow system property and storm conf have conflicts
-            if (orig_loginConfigurationFile!=null && 
-                    orig_loginConfigurationFile.length()>0 &&
-                    !loginConfigurationFile.equals(orig_loginConfigurationFile)) {
-                throw new RuntimeException("System property java.security.auth.login.config ("
-                        + orig_loginConfigurationFile
-                        +") != storm configuration java.security.auth.login.config ("
-                        + loginConfigurationFile + ")");
+            try {
+                URI config_uri = new File(loginConfigurationFile).toURI();
+                login_conf = Configuration.getInstance("JavaLoginConfig", new URIParameter(config_uri));
+            } catch (NoSuchAlgorithmException ex1) {
+                if (ex1.getCause() instanceof FileNotFoundException)
+                    throw new RuntimeException("configuration file "+loginConfigurationFile+" could not be found");
+                else throw new RuntimeException(ex1);
+            } catch (Exception ex2) {
+                throw new RuntimeException(ex2);
             }
-
-            //reset login configuration so that javax.security.auth.login will not use cache
-            Configuration.setConfiguration(null);
-
-            //use javax.security.auth.login.Configuration to obtain login configuration object
-            //login.Configuration depends on system property "java.security.auth.login.config"
-            //(see http://docs.oracle.com/javase/6/docs/jre/api/security/jaas/spec/com/sun/security/auth/login/ConfigFile.html)
-            System.setProperty("java.security.auth.login.config", loginConfigurationFile);
-            login_conf =  Configuration.getConfiguration();
-            //we reset system property to previous value if any
-            if (orig_loginConfigurationFile!=null)
-                System.setProperty("java.security.auth.login.config", orig_loginConfigurationFile);
-            else 
-                System.setProperty("java.security.auth.login.config", "");
         }
+        
         return login_conf;
     }
 
--- a/src/jvm/backtype/storm/security/auth/AuthUtils.java
+++ b/src/jvm/backtype/storm/security/auth/AuthUtils.java
@@ -4,20 +4,14 @@ import javax.security.auth.login.Configuration;
 import javax.security.auth.login.AppConfigurationEntry;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
 import backtype.storm.Config;
-import backtype.storm.utils.Utils;
-
 import java.io.IOException;
-import java.net.URL;
-import java.net.URLClassLoader;
 import java.util.Map;
 
 public class AuthUtils {
     public static final String LOGIN_CONTEXT_SERVER = "StormServer"; 
     public static final String LOGIN_CONTEXT_CLIENT = "StormClient"; 
     public static final String SERVICE = "storm_thrift_server";
-    private static final Logger LOG = LoggerFactory.getLogger(AuthUtils.class);
 
     /**
      * Construct a JAAS configuration object per storm configuration file
@@ -25,22 +19,39 @@ public class AuthUtils {
      * @return
      */
     public static synchronized Configuration GetConfiguration(Map storm_conf) {
-        Configuration.setConfiguration(null);
-
-        //exam system property first
+        //retrieve system property 
         String orig_loginConfigurationFile = System.getProperty("java.security.auth.login.config");
 
         //try to find login file from Storm configuration  
         String loginConfigurationFile = (String)storm_conf.get("java.security.auth.login.config");
-        if (loginConfigurationFile==null)
+        if ((loginConfigurationFile==null) || (loginConfigurationFile.length()==0))
             loginConfigurationFile = orig_loginConfigurationFile;
 
         Configuration login_conf = null;
         if ((loginConfigurationFile != null) && (loginConfigurationFile.length()>0)) { 
+            //We don't allow system property and storm conf have conflicts
+            if (orig_loginConfigurationFile!=null && 
+                    orig_loginConfigurationFile.length()>0 &&
+                    !loginConfigurationFile.equals(orig_loginConfigurationFile)) {
+                throw new RuntimeException("System property java.security.auth.login.config ("
+                        + orig_loginConfigurationFile
+                        +") != storm configuration java.security.auth.login.config ("
+                        + loginConfigurationFile + ")");
+            }
+
+            //reset login configuration so that javax.security.auth.login will not use cache
+            Configuration.setConfiguration(null);
+
+            //use javax.security.auth.login.Configuration to obtain login configuration object
+            //login.Configuration depends on system property "java.security.auth.login.config"
+            //(see http://docs.oracle.com/javase/6/docs/jre/api/security/jaas/spec/com/sun/security/auth/login/ConfigFile.html)
             System.setProperty("java.security.auth.login.config", loginConfigurationFile);
             login_conf =  Configuration.getConfiguration();
+            //we reset system property to previous value if any
             if (orig_loginConfigurationFile!=null)
                 System.setProperty("java.security.auth.login.config", orig_loginConfigurationFile);
+            else 
+                System.setProperty("java.security.auth.login.config", "");
         }
         return login_conf;
     }
--- a/src/jvm/backtype/storm/security/auth/IAuthorizer.java
+++ b/src/jvm/backtype/storm/security/auth/IAuthorizer.java
@@ -21,11 +21,10 @@ public interface IAuthorizer {
     
     /**
      * permit() method is invoked for each incoming Thrift request.
-     * @param contrext request context includes info about 
-     *      		   (1) remote address/subject, 
-     *                 (2) operation
-     *                 (3) configuration of targeted topology 
+     * @param context request context includes info about 
+     * @param operation operation name
+     * @param topology_storm configuration of targeted topology 
      * @return true if the request is authorized, false if reject
      */
-    public boolean permit(ReqContext context);
+    public boolean permit(ReqContext context, String operation, Map topology_conf);
 }
--- a/src/jvm/backtype/storm/security/auth/ThriftServer.java
+++ b/src/jvm/backtype/storm/security/auth/ThriftServer.java
@@ -13,7 +13,7 @@ public class ThriftServer {
     private Map _storm_conf; //storm configuration
     private TProcessor _processor = null;
     private int _port = 0;
-    private TServer _server;
+    private TServer _server = null;
     private Configuration _login_conf;
     
     public ThriftServer(Map storm_conf, TProcessor processor, int port) {
@@ -25,7 +25,7 @@ public class ThriftServer {
             //retrieve authentication configuration 
             _login_conf = AuthUtils.GetConfiguration(_storm_conf);
         } catch (Exception x) {
-            x.printStackTrace();
+            LOG.error(x.getMessage(), x);
         }
     }
 
--- a/src/jvm/backtype/storm/security/auth/ThriftServer.java
+++ b/src/jvm/backtype/storm/security/auth/ThriftServer.java
@@ -11,7 +11,7 @@ import backtype.storm.utils.Utils;
 public class ThriftServer {
     private static final Logger LOG = LoggerFactory.getLogger(ThriftServer.class);
     private Map _storm_conf; //storm configuration
-    private TProcessor _processor = null;
+    protected TProcessor _processor = null;
     private int _port = 0;
     private TServer _server = null;
     private Configuration _login_conf;
@@ -34,6 +34,15 @@ public class ThriftServer {
             _server.stop();
     }
 
+    /**
+     * Is ThriftServer listening to requests?
+     * @return
+     */
+    public boolean isServing() {
+        if (_server == null) return false;
+        return _server.isServing();
+    }
+    
     public void serve()  {
         try {
             //locate our thrift transport plugin
--- a/test/clj/backtype/storm/security/auth/auth_test.clj
+++ b/test/clj/backtype/storm/security/auth/auth_test.clj
@@ -4,8 +4,10 @@
   (:import [org.apache.thrift7 TException])
   (:import [org.apache.thrift7.transport TTransportException])
   (:import [java.nio ByteBuffer])
+  (:import [backtype.storm Config])
   (:import [backtype.storm.utils NimbusClient])
-  (:import [backtype.storm.security.auth ThriftServer ThriftClient ReqContext ReqContext$OperationType])
+  (:import [backtype.storm.security.auth AuthUtils ThriftServer ThriftClient 
+                                         ReqContext ReqContext$OperationType])
   (:use [backtype.storm bootstrap util])
   (:use [backtype.storm.daemon common])
   (:use [backtype.storm bootstrap testing])
@@ -152,15 +154,18 @@
         nimbus_client (.getClient client)]
     (log-message "(Positive authorization) Authorization plugin should accept client request")
     (.activate nimbus_client "security_auth_test_topology")
-    (.close client)))
+    (.close (.transport client))))
 
 (deftest deny-authorization-test 
   (launch-server-w-wait 6629 1000 nil
                         "backtype.storm.security.auth.authorizer.DenyAuthorizer" 
                         "backtype.storm.security.auth.SimpleTransportPlugin")
   (let [storm-conf (merge (read-storm-config)
-                           {STORM-THRIFT-TRANSPORT-PLUGIN "backtype.storm.security.auth.SimpleTransportPlugin"})
-        client (NimbusClient. storm-conf "localhost" 6629 nimbus-timeout)
+                           {STORM-THRIFT-TRANSPORT-PLUGIN "backtype.storm.security.auth.SimpleTransportPlugin"
+                           Config/NIMBUS_HOST "localhost"
+                           Config/NIMBUS_THRIFT_PORT 6629
+                           Config/NIMBUS_TASK_TIMEOUT_SECS nimbus-timeout})
+        client (NimbusClient/getConfiguredClient storm-conf)
         nimbus_client (.getClient client)]
     (log-message "(Negative authorization) Authorization plugin should reject client request")
     (is (thrown? TTransportException
@@ -208,3 +213,14 @@
            nil
            (catch TTransportException ex (.getMessage ex)))))))
 
+  (log-message "(Negative authentication) IOException")
+  (let [storm-conf (merge (read-storm-config)
+                          {STORM-THRIFT-TRANSPORT-PLUGIN "backtype.storm.security.auth.digest.DigestSaslTransportPlugin"
+                           "java.security.auth.login.config" "test/clj/backtype/storm/security/auth/jaas_digest_missing_client.conf"})]
+    (is (thrown? RuntimeException
+           (NimbusClient. storm-conf "localhost" 6630 nimbus-timeout))))
+
+(deftest test-GetTransportPlugin-throws-RuntimeException
+  (let [conf (merge (read-storm-config)
+                    {Config/STORM_THRIFT_TRANSPORT_PLUGIN "null.invalid"})]
+    (is (thrown? RuntimeException (AuthUtils/GetTransportPlugin conf nil)))))
--- a/test/clj/backtype/storm/security/auth/auth_test.clj
+++ b/test/clj/backtype/storm/security/auth/auth_test.clj
@@ -154,7 +154,7 @@
         nimbus_client (.getClient client)]
     (log-message "(Positive authorization) Authorization plugin should accept client request")
     (.activate nimbus_client "security_auth_test_topology")
-    (.close (.transport client))))
+    (.close client)))
 
 (deftest deny-authorization-test 
   (launch-server-w-wait 6629 1000 nil
--- a/test/clj/backtype/storm/security/auth/auth_test.clj
+++ b/test/clj/backtype/storm/security/auth/auth_test.clj
@@ -201,15 +201,23 @@
              (is (= "Peer indicated failure: DIGEST-MD5: cannot acquire password for unknown_user in realm : localhost" 
                     (try (NimbusClient. storm-conf "localhost" 6630 nimbus-timeout)
                       nil
-                      (catch TTransportException ex (.getMessage ex))))))))
+                      (catch TTransportException ex (.getMessage ex)))))))
 
+  (let [storm-conf (merge (read-storm-config)
+                           {STORM-THRIFT-TRANSPORT-PLUGIN "backtype.storm.security.auth.digest.DigestSaslTransportPlugin"
+                            "java.security.auth.login.config" "test/clj/backtype/storm/security/auth/nonexistent.conf"})]
+        (testing "(Negative authentication) nonexistent configuration file"
+                 (is (thrown? RuntimeException 
+                              (NimbusClient. storm-conf "localhost" 6630 nimbus-timeout)))))
+  
   (let [storm-conf (merge (read-storm-config)
                           {STORM-THRIFT-TRANSPORT-PLUGIN "backtype.storm.security.auth.digest.DigestSaslTransportPlugin"
                            "java.security.auth.login.config" "test/clj/backtype/storm/security/auth/jaas_digest_missing_client.conf"})]
-    (testing "(Negative authentication) IOException"
+    (testing "(Negative authentication) Missing client"
              (is (thrown? RuntimeException
-                          (NimbusClient. storm-conf "localhost" 6630 nimbus-timeout)))))
-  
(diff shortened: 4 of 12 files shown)


src/jvm/backtype/storm/security/auth/IAuthorization.java
```diff
@@ -0,0 +1,26 @@
...
+ * Nimbus could be configured with an authorization plugin.
+ * If not specified, all requests are authorized.
+ * 
+ * You could specify the authorization plugin via storm parameter. For example:
+ *  storm -c nimbus.authorization.classname=backtype.storm.security.auth.DefaultAuthorizer ...
+ *  
+ * You could also specify it via storm.yaml:
+ *   nimbus.authorization.classname: backtype.storm.security.auth.DefaultAuthorizer
+ *   
+ * @author afeng
```
nathanmarz: Please remove @author tags from contributions

nathanmarz:
I don't understand a lot of this pull request – that's mostly my own ignorance of authentication/authorization and the libraries you're using. Can you explain the workflow of how this stuff will work once it's integrated into Nimbus and the client?

Also, what is the purpose of all the files you put in conf/ ? Are those all required to get this to work? Or are those just examples of different, independent auth mechanisms?

Also – is it possible to add tests for all this? Or will the tests have to come once it's integrated with Nimbus and the clients?

anfeng:
The conf files are examples of login configurations. They show how Digest
and Kerberos authentication should be configured.

Let me give a  high-level overview of request flow from Nimbus Client to
Nimbus Server.
1. Nimbu server (ThriftServer) will be configured with a login plugin and an
authorization plugin.

> - Ex., storm -c 
>   nimbus.authorization.classname=backtype.storm.security.auth.NoopAuthorizer  -c
>   java.security.auth.login.config=jaas_kerberos_cluster.conf nimbus
> - If login configuration is not given, we assume AnonymousAuthentication.
> - If authorization plugin is not given, we assume all requests should be
>   processed.
> - Nimbus client (ThriftClient) will be configured with a login plugin
> - Ex., storm -c java.security.auth.login.config=jaas_kerberos_cluster.conf
>   supervisor
> - Ex., storm -c java.security.auth.login.config=jaas_kerberos_launcher.conf
>   jar
> - If login configuration is not given, we assume AnonymousAuthentication.
> - When Nimbus client is connecting to Nimbus Server, SASL negotiation will
>   occur. The SASL negotiation will invoke the authentication callback included
>   in login configuration
> - Ex., SaslClientCallbackHandler and SaslServerCallbackHandler are the
>   callback class for digest/kerberos authentication
> - Ex., AnonymousAuthenticationProvider contains the callback class for
>   anonymous mechanism (no auth)
> - If authentication failed at step 3, Nimbus client will get an exception
>   indicating authentication failure.
> - If authentication is successful, ThriftServer::SaslProcessor::process()
>   method will be invoked.
> - ThriftServer::SaslProcessor::process() will create a ReqContext object
>   with info such as remote address, remote principal (User) and operation.
> - Authorization plugin will be invoked with a newly created ReqContext.
>   Authorization plugin has a method permit(ReqContext context) , which return
>   true if and only if authorized.
> - Ex., NoopAuthorizer always return true. It logs the access into access log
>   (see logback configuration)
> - Ex., DenyAuthorizer.java always return false (reject)
> - If authorization is rejected, Nimbus client will get an exception for
>   authorization failure.
> - If authorization is successful, Nimbus will process the request Š

anfeng:
Nathan, please check out the newly added test case for authentication and authroization. 
https://github.com/anfeng/storm/blob/master-auth/test/clj/backtype/storm/security/auth/auth_test.clj

The test cases covers anonymous/digest authentication, and authorizations.
It will be challenging to include Kerberos in test cases.

Here is the log from test cases.

Testing backtype.storm.security.auth.auth-test
1557 [main] INFO  backtype.storm.security.auth.auth-test - Waiting for Nimbus Server...
1633 [clojure-agent-send-off-pool-0] INFO  backtype.storm.security.auth.auth-test - Starting Nimbus server...
1637 [clojure-agent-send-off-pool-0] INFO  backtype.storm.security.auth.ThriftServer - Starting SASL ANONYMOUS server at port:6627
2559 [main] INFO  backtype.storm.security.auth.auth-test - Starting Nimbus client w/ anonymous authentication
2571 [main] INFO  backtype.storm.security.auth.ThriftClient - SASL ANONYMOUS client transport is being established
2584 [pool-2-thread-1] INFO  backtype.storm.security.auth.DenyAuthorizer - Access  from: /127.0.0.1 principal:anonymous op:ACTIVATE_TOPOLOGY topoology:bogus_topology
2587 [pool-2-thread-1] ERROR org.apache.thrift7.server.TThreadPoolServer - Error occurred during processing of message.
java.lang.RuntimeException: ACTIVATE_TOPOLOGY on topology bogus_topology is not authorized
    at backtype.storm.security.auth.auth_test$check_authorization_BANG_.invoke(auth_test.clj:58) ~[na:na]
    at backtype.storm.security.auth.auth_test$dummy_service_handler$reify__487.activate(auth_test.clj:78) ~[na:na]
    at backtype.storm.generated.Nimbus$Processor$activate.getResult(Nimbus.java:1307) ~[classes/:na]
    at backtype.storm.generated.Nimbus$Processor$activate.getResult(Nimbus.java:1295) ~[classes/:na]
    at org.apache.thrift7.ProcessFunction.process(ProcessFunction.java:32) ~[libthrift7-0.7.0.jar:0.7.0]
    at org.apache.thrift7.TBaseProcessor.process(TBaseProcessor.java:34) ~[libthrift7-0.7.0.jar:0.7.0]
    at backtype.storm.security.auth.ThriftServer$SaslProcessor.process(ThriftServer.java:180) ~[classes/:na]
    at org.apache.thrift7.server.TThreadPoolServer$WorkerProcess.run(TThreadPoolServer.java:176) ~[libthrift7-0.7.0.jar:0.7.0]
    at java.util.concurrent.ThreadPoolExecutor$Worker.runTask(ThreadPoolExecutor.java:886) [na:1.6.0_37]
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:908) [na:1.6.0_37]
    at java.lang.Thread.run(Thread.java:680) [na:1.6.0_37]

2589 [main] INFO  backtype.storm.security.auth.auth-test - Waiting for Nimbus Server...
2602 [clojure-agent-send-off-pool-1] INFO  backtype.storm.security.auth.auth-test - Starting Nimbus server...
2609 [clojure-agent-send-off-pool-1] INFO  backtype.storm.security.auth.ThriftServer - Starting DIGEST server at port:6628
3590 [main] INFO  backtype.storm.security.auth.auth-test - Starting Nimbus client w/ anonymous authentication (expect authentication failure
3600 [main] INFO  backtype.storm.security.auth.ThriftClient - SASL ANONYMOUS client transport is being established
3602 [main] ERROR backtype.storm.security.auth.ThriftClient - Peer indicated failure: Unsupported mechanism type ANONYMOUS
3602 [pool-3-thread-1] ERROR org.apache.thrift7.server.TThreadPoolServer - Error occurred during processing of message.
java.lang.RuntimeException: org.apache.thrift7.transport.TTransportException: Unsupported mechanism type ANONYMOUS
    at org.apache.thrift7.transport.TSaslServerTransport$Factory.getTransport(TSaslServerTransport.java:218) ~[libthrift7-0.7.0.jar:0.7.0]
    at org.apache.thrift7.server.TThreadPoolServer$WorkerProcess.run(TThreadPoolServer.java:170) ~[libthrift7-0.7.0.jar:0.7.0]
    at java.util.concurrent.ThreadPoolExecutor$Worker.runTask(ThreadPoolExecutor.java:886) [na:1.6.0_37]
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:908) [na:1.6.0_37]
    at java.lang.Thread.run(Thread.java:680) [na:1.6.0_37]
Caused by: org.apache.thrift7.transport.TTransportException: Unsupported mechanism type ANONYMOUS
    at org.apache.thrift7.transport.TSaslTransport.sendAndThrowMessage(TSaslTransport.java:221) ~[libthrift7-0.7.0.jar:0.7.0]
    at org.apache.thrift7.transport.TSaslServerTransport.handleSaslStartMessage(TSaslServerTransport.java:137) ~[libthrift7-0.7.0.jar:0.7.0]
    at org.apache.thrift7.transport.TSaslTransport.open(TSaslTransport.java:253) ~[libthrift7-0.7.0.jar:0.7.0]
    at org.apache.thrift7.transport.TSaslServerTransport.open(TSaslServerTransport.java:40) ~[libthrift7-0.7.0.jar:0.7.0]
    at org.apache.thrift7.transport.TSaslServerTransport$Factory.getTransport(TSaslServerTransport.java:215) ~[libthrift7-0.7.0.jar:0.7.0]
    ... 4 common frames omitted
3602 [main] INFO  backtype.storm.security.auth.auth-test - Starting Nimbus client w/ digest authentication (expect authentication success)
3613 [pool-3-thread-2] INFO  backtype.storm.security.auth.NoopAuthorizer - bob op:ACTIVATE_TOPOLOGY topoology:bogus_topology

Ran 10 tests containing 33 assertions.
0 failures, 0 errors.

src/jvm/backtype/storm/security/auth/ThriftClient.java
```diff
@@ -0,0 +1,153 @@
...
+
+			String loginConfigurationFile = System.getProperty("java.security.auth.login.config");
+			if ((loginConfigurationFile==null) || (loginConfigurationFile.length()==0)) {
+				//apply Storm configuration for JAAS login 
+				Map conf = Utils.readStormConfig();
+				loginConfigurationFile = (String)conf.get("java.security.auth.login.config");
+			}
+			if ((loginConfigurationFile==null) || (loginConfigurationFile.length()==0)) { //ANONYMOUS
+				LOG.info("SASL ANONYMOUS client transport is being established");
+				_transport = new TSaslClientTransport(AuthUtils.ANONYMOUS,
```
nathanmarz: What is this? I thought in the case of not configuring security that everything would work exactly as it does now. Why is this necessary? Why not just use a regular framed transport in this case?

src/jvm/backtype/storm/security/auth/ThriftServer.java
```diff
@@ -0,0 +1,251 @@
...
+
+		try {
+			TSaslServerTransport.Factory factory = new TSaslServerTransport.Factory();
+			serverTransport = new TServerSocket(_port);
+
+			if ((_loginConfigurationFile==null) || (_loginConfigurationFile.length()==0)) { //ANONYMOUS
+				factory.addServerDefinition(AuthUtils.ANONYMOUS, AuthUtils.SERVICE, "localhost", null, null);			
+
+				LOG.info("Starting SASL ANONYMOUS server at port:" + _port);
+				_server = new TThreadPoolServer(new TThreadPoolServer.Args(serverTransport).
```
nathanmarz: Why are you doing different kind of Thrift server than Nimbus currently uses? Stick with the HsHa server.

src/jvm/backtype/storm/security/auth/ThriftServer.java
```diff
@@ -0,0 +1,251 @@
...
+			} else {	
+				//retrieve authentication configuration from java.security.auth.login.config
+				Configuration auth_conf =  AuthUtils.getConfiguration(_loginConfigurationFile);
+
+				//login our user
+				CallbackHandler auth_callback_handler = new SaslServerCallbackHandler(auth_conf);
+				Login login = new Login(AuthUtils.LoginContextServer, auth_callback_handler);
+				Subject subject = login.getSubject();
+
+				if (!subject.getPrivateCredentials(KerberosTicket.class).isEmpty()) { //KERBEROS
```
nathanmarz: This doesn't seem very pluggable.

nathanmarz:
A few comments:
1. Can you update your editor to not insert tabs? It's much cleaner and cross-editor compatible to use spaces. For Java code I recommend using 4 spaces instead of a tab (most editors can do this for you automatically)
2. Can you add a comment inside the conf/ files you added indicating that these are examples and when a user would want to use them?
3. I don't like how there's all this new code in effect even in the case that no auth is configured. In that case, all the security stuff should not execute and Nimbus and the client should operate exactly as they do today.
4. The authentication mechanisms seem to be hardcoded. I would much rather see this stuff done via a plugin model. Actually – if we could make it pluggable in a simple way and then include the specific types of authorization as separate modules to Storm itself, that would be ideal. I'm thinking there would be a config like "storm.authentication" that points to some class that implements some interface with the extension points for authentication. Based on the pull request, the extensions points would have to be able to influence the transport used and some of the logic when Nimbus receives request, among other things.

anfeng:
Please take a look at the latest revision. It should address all your concerns. 

We introduced a new configuration parameter "storm.thrift.transport", which should refer to the class name of our transporter plugin. In defaults.yaml, we have added the following line:
  storm.thrift.transport: "backtype.storm.security.auth.SimpleTransportPlugin"

Transporter plugins need to implement an interface, called ITransportPlugin, which has 2 methods:

```
/* Create a server associated with a given port and service handler */
public TServer getServer(int port, TProcessor processor);

/* Connect to the specified server via framed transport */
public TTransport connect(TTransport transport, String serverHost) 
```

We have 4 built-in plugins: 
 (1) Simple plugin (default) ... compatible w/ existing Thrift protocol
 (2) Digest/Kerberos/Anonymous ... All these are based on Thrift SASL

auth_test.clj has been revised to verify plugins: Simple, Digest and Anonymous.

anfeng:
Nathan, please review the changes. I will hold off Kerberos plugin to a separated pull request. This pull request is focused on the authentication/authorization framework.

src/jvm/backtype/storm/security/auth/AuthUtils.java
```diff
@@ -0,0 +1,81 @@
...
+    public static final String LOGIN_CONTEXT_CLIENT = "StormClient"; 
+    public static final String SERVICE = "storm_thrift_server";
+    private static final Logger LOG = LoggerFactory.getLogger(AuthUtils.class);
+
+    /**
+     * Construct a JAAS configuration object per storm configuration file
+     * @param storm_conf Storm configuration 
+     * @return
+     */
+    public static synchronized Configuration GetConfiguration(Map storm_conf) {
```
nathanmarz: Why is this method synchronized? And also, why does it have side effects (setting properties and whatnot)? Would it make more sense to put this initialization code in the main method of nimbus (and possibly also supervisor and/or worker as necessary)?

src/jvm/backtype/storm/security/auth/IAuthorizer.java
```diff
@@ -0,0 +1,23 @@
...
+ * Nimbus could be configured with an authorization plugin.
+ * If not specified, all requests are authorized.
+ * 
+ * You could specify the authorization plugin via storm parameter. For example:
+ *  storm -c nimbus.authorization.class=backtype.storm.security.auth.NoopAuthorizer ...
+ *  
+ * You could also specify it via storm.yaml:
+ *   nimbus.authorization.class: backtype.storm.security.auth.NoopAuthorizer
+ */
+public interface IAuthorizer {
```
nathanmarz, line 15: Let's add a prepare(Map conf) method.

src/jvm/backtype/storm/security/auth/ITransportPlugin.java
```diff
@@ -0,0 +1,30 @@
...
+import org.apache.thrift7.transport.TTransport;
+import org.apache.thrift7.transport.TTransportException;
+
+/**
+ * Interface for Thrift Transport plugin
+ * 
+ * Each plugin should have a constructor 
+ *  Foo(Configuration login_conf)
+ */
+public interface ITransportPlugin {
```
nathanmarz, line 16: Rather than mandate that type of constructor, let's add a prepare(Map conf) method. Also, what "Configuration" class are you referring to in that comment?

src/jvm/backtype/storm/security/auth/SimpleTransportPlugin.java
```diff
@@ -0,0 +1,113 @@
...
+                req_context.setRemoteAddress(socket.getInetAddress());                
+            } 
+
+            //anonymous user
+            req_context.setSubject(null);
+
+            //invoke service handler
+            try {
+                return wrapped.process(inProt, outProt);
+            } catch (RuntimeException ex) {
```
nathanmarz: Why are you suppressing exception here? Let the runtime exception bubble up. This is changing the behavior of the wrapped processor.

src/jvm/backtype/storm/security/auth/ThriftServer.java
```diff
@@ -0,0 +1,53 @@
...
+
+            //server
+            _server = transportPlugin.getServer(_port, _processor);
+
+            //start accepting requests
+            _server.serve();
+        } catch (Exception ex) {
+            LOG.error("ThriftServer is being stopped due to: " + ex, ex);
+            if (_server != null) _server.stop();
+            System.exit(1); //shutdown server process since we could not handle Thrift requests any more
```
nathanmarz: Should use System.halt and not System.exit. Nimbus is fail fast and so should aggressively suicide in case it hits a bad situation. System.exit can sometimes freeze without exiting the process when the process has gotten seriously messed up (like memory issues).

nathanmarz:
I made a bunch more inline comments. Finally, one other issue that's pervasive throughout the pull request is logging in the catch block of an exception before rethrowing the exception. Unless you have extra information to log besides what's in the exception, that's completely redundant. So you should just rethrow the exception as a runtime exception without doing any additional logging. Something higher up in the stack will log it.

src/jvm/backtype/storm/security/auth/AuthUtils.java
anfeng: This method is synchronized since javax.security.auth.login.Configuration:getConfiguration() depends on system property "java.security.auth.login.config". We need to make sure that the login configuration object is based on the given storm configuration, and is not impacted due to 2 threads w/ different storm conf files.

src/jvm/backtype/storm/security/auth/IAuthorizer.java
anfeng, line 15: agree

src/jvm/backtype/storm/security/auth/ITransportPlugin.java
anfeng, line 16: How about the following method? The 2nd parameter is driven from storm_conf. We add it here to make plugin development easier, and ensure all plugins use "java.security.auth.login.config" in storm configuration.  

```
/**
 * Invoked once immediately after construction
 * @param storm_conf Storm configuration 
 * @param login_conf login configuration
 */
void prepare(Map storm_conf, javax.security.auth.login.Configuration login_conf);
```

src/jvm/backtype/storm/security/auth/SimpleTransportPlugin.java
anfeng: revised

test/clj/backtype/storm/security/auth/auth_test.clj
```diff
@@ -0,0 +1,226 @@
...
+                     NIMBUS-THRIFT-PORT server-port
+                     STORM-THRIFT-TRANSPORT-PLUGIN transportPluginClass})
+        conf (if login-cfg (merge conf1 {"java.security.auth.login.config" login-cfg}) conf1)
+        nimbus (nimbus/standalone-nimbus)
+        service-handler (dummy-service-handler conf nimbus)
+        server (ThriftServer. conf (Nimbus$Processor. service-handler) (int (conf NIMBUS-THRIFT-PORT)))]
+    (.addShutdownHook (Runtime/getRuntime) (Thread. (fn [] (.stop server))))
+    (.serve server)))
+
+(defn launch-server-w-wait [server-port ms login-cfg aznClass transportPluginClass]
```
nathanmarz: Tests cannot rely on sleeping to work correctly. This needs to be changed to only return when the Nimbus server is ready to receive requests.

test/clj/backtype/storm/security/auth/auth_test.clj
```diff
@@ -0,0 +1,226 @@
...
+  (.start (Thread. #(launch-test-server server-port login-cfg aznClass transportPluginClass)))
+  (Thread/sleep ms))
+
+(deftest Simple-authentication-test 
+  (launch-server-w-wait 6627 1000 nil nil "backtype.storm.security.auth.SimpleTransportPlugin")
+
+  (log-message "(Positive authentication) Server and Client with simple transport, no authentication")
```
nathanmarz: Let's get rid of all this logging in the tests. If you want to mark what each portion is testing, use the "testing" form. ( http://clojuredocs.org/clojure_core/clojure.test/testing )

test/clj/backtype/storm/security/auth/auth_test.clj
```diff
@@ -0,0 +1,226 @@
...
+     :heartbeats-cache (atom {})
+     :downloaders nil
+     :uploaders nil
+     :uptime (uptime-computer)
+     :validator nil
+     :timer nil
+     :scheduler nil
+     }))
+
+(defn update-req-context! [nimbus storm-name storm-conf operation]
```
nathanmarz: Why is this code necessary in the tests? Isn't the ReqContext supposed to be managed automatically by the underlying Thrift components (as you do in TUGIWrapProcessor)? If not, then I think the ReqContext should only contain the stuff that can be filled in automatically by the underlying Thrift transport, and the rest of the information should be passed to the authorizer as another immutable argument.

src/jvm/backtype/storm/security/auth/AuthUtils.java
```diff
@@ -0,0 +1,81 @@
...
+        Configuration.setConfiguration(null);
+
+        //exam system property first
+        String orig_loginConfigurationFile = System.getProperty("java.security.auth.login.config");
```
nathanmarz: Why are there two ways to configure? I'd much rather there only be one way, through the storm.yaml.

src/jvm/backtype/storm/security/auth/AuthUtils.java
```diff
@@ -0,0 +1,81 @@
...
+
+        //try to find login file from Storm configuration  
+        String loginConfigurationFile = (String)storm_conf.get("java.security.auth.login.config");
+        if (loginConfigurationFile==null)
+            loginConfigurationFile = orig_loginConfigurationFile;
+
+        Configuration login_conf = null;
+        if ((loginConfigurationFile != null) && (loginConfigurationFile.length()>0)) { 
+            System.setProperty("java.security.auth.login.config", loginConfigurationFile);
```
nathanmarz: What's the purpose of all this code? Why do you need to set these properties? Why do you set the property and then potentially override it yet again in the next line?

nathanmarz:
Sorry for the delay. I just left some more comments. This is the last round of comments/changes before I'm ready to merge in. 

Once we're done with this, I'm going to restructure Storm to have a multi-module setup, and we can then put most of the specific protocol implementations in the security package. This will make it much easier to manage the codebase, as I think it's getting unwieldy now. Likewise, I'm going to put Trident into its own module.

src/jvm/backtype/storm/security/auth/AuthUtils.java
anfeng: Nathan, JAAS configuration is wide used industry standard (http://docs.oracle.com/javase/1.5.0/docs/guide/security/jaas/tutorials/LoginConfigFile.html), and we should avoid invent a new wheel here. For example, ZK is taking a similar approach, see http://hbase.apache.org/book/zk.sasl.auth.html.
The standard JAAS configuration also enables us to use JDK lib for authentication (javax.security.auth.*).

nathanmarz: Ah ok, I see. In that case, I think that if they're both set and are not equal, an error should be thrown.

src/jvm/backtype/storm/security/auth/AuthUtils.java
anfeng: The default implementation for javax.security.auth.login.Configuration is dependent on system property java.security.auth.login.config. see http://docs.oracle.com/javase/6/docs/jre/api/security/jaas/spec/com/sun/security/auth/login/ConfigFile.html

AuthUtils:GetConfiguration() is trying to the storm config to obtain login Configuration object. The last setProperty() is try to bring system property to previous state.

anfeng:
Nathan, I just uploaded changes to address your questions/comments. Please let me know if you have comments. 

It will be great to have modules in Storm. It will be good to enable users to select only modules per their needs.

src/jvm/backtype/storm/security/auth/AuthUtils.java
nathanmarz: I'm still uncomfortable with this code for two reasons:
1. "is not impacted due to 2 threads w/ different storm conf files." This is not true with the current code, as if the system property is not set, you set the property on line 48. So the second thread with the differing storm conf would be impacted, and if it had a differing java.security.auth.login.config in the storm conf, then the second thread would have a runtime exception. 
2. The link you sent says "The default Configuration implementation from Sun Microsystems reads configuration information from configuration files". Will this code work appropriately on other JVM implementations?

nathanmarz:
Thanks for addressing those problems. I still have concerns on the AuthUtils.getConfiguration(Map) method and left another comment on that method.

src/jvm/backtype/storm/security/auth/AuthUtils.java
anfeng: For your statement (1), we address the issue via "sunchronized" in the method. However, we could slightly improved the code by using storm configuration only. See the code below.

The code only work if the default JAAS Configuration Provider is used. We could avoid such a limitation
by provide our own implementation of javax.security.auth.login.Configuration. The implementation will essentially be a copy of sun's ConfigFile.java (see http://www.docjar.com/html/api/com/sun/security/auth/login/ConfigFile.java.html).  Should we do that?

```
public static synchronized Configuration GetConfiguration(Map storm_conf) {
    //try to find login file from Storm configuration  
    String loginConfigurationFile = (String)storm_conf.get("java.security.auth.login.config");

    Configuration login_conf = null;
    if ((loginConfigurationFile != null) && (loginConfigurationFile.length()>0)) { 
        //retrieve system property 
        String orig_loginConfigurationFile = System.getProperty("java.security.auth.login.config");

        //use javax.security.auth.login.Configuration to obtain login configuration object
        //login.Configuration depends on system property "java.security.auth.login.config"
        //(see http://docs.oracle.com/javase/6/docs/jre/api/security/jaas/spec/com/sun/security/auth/login/ConfigFile.html)
        System.setProperty("java.security.auth.login.config", loginConfigurationFile);
        Configuration.setConfiguration(null);
        login_conf =  Configuration.getConfiguration();

        //we reset system property to previous value if any
        if (orig_loginConfigurationFile!=null)
            System.setProperty("java.security.auth.login.config", orig_loginConfigurationFile);
        else 
            System.setProperty("java.security.auth.login.config", "");
    }
    return login_conf;
}
```

anfeng: I think that we should live with the limitation of default Configuration Provider. Many other projects took that route: Hadoop, Zookeeper etc. 

Also, Sun's ConfigFile is under GPL v2 license, which is not a preferred license for various enterprise.

nathanmarz: Oh sorry, you're right. I read the code wrong. I didn't realize that you reset the value of the system property at the end. Though I do agree the code is easier to understand by only using what's in the storm conf, and it's arguably better to only have one way to configure this.

nathanmarz: I think we can improve this code by being explicit about using the file config implementation of Configuration. This will also remove the need to do any mucking around with System properties. Let's use Configuration#getInstance and set the parameters according to http://docs.oracle.com/javase/6/docs/technotes/guides/security/StandardNames.html#Configuration and http://docs.oracle.com/javase/6/docs/api/javax/security/auth/login/Configuration.html#getInstance(java.lang.String, javax.security.auth.login.Configuration.Parameters)

So it should look like: `Configuration.getInstance("JavaLoginConfig", new URIParameter(new URI(loginConfigurationFile)))`

This has the benefits of:
1. Being explicit that we're using a config file
2. Throwing an exception if the JVM doesn't have the config file implementation
3. No side effects, so it's a pure function and we can get rid of the `synchronized`

Does that make sense?

anfeng: Outstanding suggestion. I just made the change in AuthUtils accordingly, and
also add 1 test case.
This is much cleaner. Thanks!

From:  Nathan Marz notifications@github.com
Reply-To:  nathanmarz/storm
<reply+p-3311003-ef3b34803e518ea80beab38a67b71f158f45f989-2708344@reply.gith
ub.com>
Date:  Sunday, March 10, 2013 3:07 PM
To:  nathanmarz/storm storm@noreply.github.com
Cc:  Andy Feng andy.feng@gmail.com
Subject:  Re: [storm] Apply Thrift SASL client/server framework for
authentication/authorization/audit (#469)

In src/jvm/backtype/storm/security/auth/AuthUtils.java:

> +import java.net.URL;
> +import java.net.URLClassLoader;
> +import java.util.Map;
> +
> +public class AuthUtils {
> -    public static final String LOGIN_CONTEXT_SERVER = "StormServer";
> -    public static final String LOGIN_CONTEXT_CLIENT = "StormClient";
> -    public static final String SERVICE = "storm_thrift_server";
> -    private static final Logger LOG =
>   LoggerFactory.getLogger(AuthUtils.class);
>   +
> -    /**
> -     \* Construct a JAAS configuration object per storm configuration file
> -     \* @param storm_conf Storm configuration
> -     \* @return
> -     */
> -    public static synchronized Configuration GetConfiguration(Map storm_conf)
>   {
>   I think we can improve this code by being explicit about using the file
>   config implementation of Configuration. This will also remove the need to do
>   any mucking around with System properties. Let's use
>   Configuration#getInstance and set the parameters according to
>   http://docs.oracle.com/javase/6/docs/technotes/guides/security/StandardNames
>   .html#Configuration and
>   http://docs.oracle.com/javase/6/docs/api/javax/security/auth/login/Configura
>   tion.html#getInstance(java.lang.String,
>   javax.security.auth.login.Configuration.Parameters)

So it should look like: Configuration.getInstance("JavaLoginConfig", new
URIParameter(new URI(loginConfigurationFile)))

This has the benefits of:
1. Being explicit that we're using a config file
2. Throwing an exception if the JVM doesn't have the config file
implementation
3. No side effects, so it's a pure function and we can get rid of the
synchronized
Does that make sense?

‹
Reply to this email directly or view it on GitHub
https://github.com/nathanmarz/storm/pull/469/files#r3311003 .

nathanmarz:
I tried merging it in, but the tests don't pass. In particular, the test at line auth_test#201 failed. I'm not a fan of testing against the string of an exception, as that's extremely brittle. Can you change it to instead check that the exception cause is of the appropriate type? I just committed a little helper that makes that extremely easy: backtype.storm.util#thrown-cause? . Use it like so:

(is (thrown-cause? SocketTimeoutException (NimbusClient. <rest of code here>)))

thrown-case? walks the exception cause stack to see if an exception of that type is found.

anfeng:
I made changes to auth_test to use thrown_cause? whenever possible. All string comparisons have been replaced by thrown? or thrown-cause?.

nathanmarz:
Merged! Made a couple of small changes on my own:
1. Update auth_test to shut down the server at the end of the test. I wrapped launch-server and (.stop server) in a simple macro to do this.
2. Removed the usage of NIMBUS_TASK_TIMEOUT_SECS for specifying the NimbusClient timeout. If you want that configurable, that should be its own config and you can just submit that in another pull request.

PR #470, opened 2013-02-13
Disallow XML characters in topology names

d2r:
Certain characters in topology names are parsed as XML in the browser
when served via the UI.


nathanmarz:
You should remove e2b3a0c from this pull request, as that's for a separate issue. It's easiest to make pull requests from their own branches so that they're kept independent.

Wouldn't it be better to just fix the UI so that it can render any topology name correctly?

PR #467, opened 2013-02-12
upgrade zookeeper to 3.4.5

anfeng:
Nathan, let's upgrade Zookeeper to 3.4.5 (the latest stable release).

PR #471, opened 2013-02-13
use exponential backoff retry policy

d2r:
Use com.netflix.curator.retry.ExponentialBackoffRetry instead of com.netflix.curator.retry.RetryNTimes for ZK connections.


nathanmarz:
Did you run into something in particular that required this?

d2r:
Reiterating Andy's rationale for the change:

With a large cluster, after Zookeeper is not accessible for a period of time, we found Zookeeper can get overwhelmed when trying to recover.  The exponential back-off helps in those situations.

src/jvm/backtype/storm/utils/Utils.java
```diff
@@ -299,7 +299,7 @@ public static CuratorFramework newCurator(Map conf, List<String> servers, Object
                     .connectString(zkStr)
                     .connectionTimeoutMs(Utils.getInt(conf.get(Config.STORM_ZOOKEEPER_CONNECTION_TIMEOUT)))
                     .sessionTimeoutMs(Utils.getInt(conf.get(Config.STORM_ZOOKEEPER_SESSION_TIMEOUT)))
-                    .retryPolicy(new RetryNTimes(Utils.getInt(conf.get(Config.STORM_ZOOKEEPER_RETRY_TIMES)), Utils.getInt(conf.get(Config.STORM_ZOOKEEPER_RETRY_INTERVAL))));
+                    .retryPolicy(new ExponentialBackoffRetry(Utils.getInt(conf.get(Config.STORM_ZOOKEEPER_RETRY_TIMES)), Utils.getInt(conf.get(Config.STORM_ZOOKEEPER_RETRY_INTERVAL))));
```
nathanmarz: According to http://netflix.github.com/curator/doc/com/netflix/curator/retry/ExponentialBackoffRetry.html, these arguments are reversed here.

nathanmarz:
This is a good idea in principle. Some questions:
1. What do the time intervals look like exactly? If your base is 100ms, what will the next 10 tries look like? 200ms, 400ms, 800ms, 1.6s, 3.2s, etc?
2. Is there a way to set a ceiling on the retry time? Is this even desirable? Seems like some serious problems could arise if a worker is waiting 30s to retry an operation – like for example, a worker timeout.

d2r:
1. They would look like: 100~200ms, 100~400ms, 100~800ms, etc.  With a base of 100ms, the interval for a given `retryCount` is random between 100ms ~ 2^(`retryCount` + 1).  So it is randomized, with the upper limit increasing exponentially.
2. Yes we can, and I agree that probably would be a good idea. It defaults to `Integer.MAX_VALUE`.  What would be a good ceiling?

d2r:
Further notes on the ceiling.  The version of curator we pull in now (1.0.1) does not have a ceiling.  We'd need at least 1.2.6.

Also, in the curator git revision comments there is a note that they changed the ceiling from 60000ms to `Integer.MAX_VALUE` to avoid "breaking existing code."

d2r:
What we really want is BoundedExponentialBackoffRetry, which was added in curator 1.2.5.

But I note that there is a significant hurdle to pulling in a newer curator: #225, especially it's dependency on zookeeper 3.4.5 #467

d2r:
Put a configurable ceiling on the backoff retry, defaults to 30s, but we can discuss what it should be.
Added JUnit and a test for the change.

Trying to avoid pulling in a new Curator.  We may not want to go down this road either...

nathanmarz:
Yea, we've had some trouble with the new ZK client library that we're trying to hash out. 

As for the diff – let's keep the tests in Clojure. Just add a new util_test.clj file into test/clj/backtype/storm/

d2r:
Sure, I can rewrite the test. 

Curious, what is the barrier to using JUnit tests?

nathanmarz:
JUnit adds no additional functionality compared to what we can already do with clojure.test, so it's simpler to stick with the existing unit testing framework. This also reduces dependencies of the project and makes it simpler to run the unit tests (just lein test).

d2r:
OK, rewritten with a clojure test.

test/clj/backtype/storm/util_test.clj
```diff
@@ -0,0 +1,33 @@
...
+        servers ["bogus_server"]
+        arbitrary_port 42
+        curator (Utils/newCurator conf servers arbitrary_port)
+        retry (-> curator .getZookeeperClient .getRetryPolicy)
+       ]
+    (is (.isAssignableFrom ExponentialBackoffRetry (.getClass retry)))
+    (is (= (.getBaseSleepTimeMs retry) expected_interval)) 
+    (is (= (.getN retry) expected_retries)) 
+    (is (= (.getMaxRetryInterval retry) expected_ceiling)) 
+    ; It would be very unlikely for this to fail three times.
```
nathanmarz: What does this mean exactly? Isn't the function stateless, so can't you just test the value once?

nathanmarz:
OK – I think we're close to finishing this. First off, let's get rid of the project.clj changes as those are no longer necessary. I also made one more comment in the code,

test/clj/backtype/storm/util_test.clj
d2r: The result is psuedo-randomly distributed, so there is a slim chance with these values that working code could trip the assertion.

I could trivially set the ceiling to be less than the base, and that would guarantee the assertion is true.

d2r:
Reverted the project.clj change, sorry about that.  Replied to your in-line comment.

nathanmarz:
Merged. I changed the config to explicitly say millis.

PR #473, opened 2013-02-13
Add test for BlowfishTupleSerializer

d2r:
This complements 01c4147 from pull request 461.

### DIFF
--- a/test/clj/backtype/storm/security/serialization/BlowfishTupleSerializer_test.clj
+++ b/test/clj/backtype/storm/security/serialization/BlowfishTupleSerializer_test.clj
@@ -1,6 +1,7 @@
 (ns backtype.storm.security.serialization.BlowfishTupleSerializer-test
   (:use [clojure test]
         [backtype.storm.util :only (exception-cause?)]
+        [clojure.string :only (join split)]
   )
   (:import [backtype.storm.security.serialization BlowfishTupleSerializer]
            [backtype.storm.utils ListDelegate]
@@ -10,18 +11,10 @@
 )
 
 (deftest test-constructor-throws-on-null-key
-  (let [conf {}]
-    (try
-      (new BlowfishTupleSerializer nil conf)
-      (catch Throwable t
-        (is (exception-cause? java.lang.RuntimeException t)
-        "Throws RuntimeException when no encryption key is given.")
-      )
-    )
-  )
+  (is (thrown? RuntimeException (new BlowfishTupleSerializer nil {}))
+      "Throws RuntimeException when no encryption key is given.")
 )
 
-(use '[clojure.string :only (join split)])
 (deftest test-encrypts-and-decrypts-message
 
   (let [
@@ -50,7 +43,7 @@
        ]
     (-> delegate (.addAll strlist))
     (-> writer-bts (.write kryo output delegate))
-    (-> input (.setBuffer (-> output (.getBuffer))))
+    (.setBuffer input (.getBuffer output))
     (is 
       (=
         test-text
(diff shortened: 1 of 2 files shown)


test/clj/backtype/storm/security/serialization/BlowfishTupleSerializer_test.clj
```diff
@@ -0,0 +1,70 @@
...
+  (:import [backtype.storm.security.serialization BlowfishTupleSerializer]
+           [backtype.storm.utils ListDelegate]
+           [com.esotericsoftware.kryo Kryo]
+           [com.esotericsoftware.kryo.io Input Output]
+  )
+)
+
+; Exceptions are getting wrapped in RuntimeException.  This might be due to
+; CLJ-855.
+(defn- unpack-runtime-exception [expression]
```
nathanmarz: You should use backtype.storm.util/exception-cause? rather than thrown?. It walks the cause stack checking for an exception type of that class.

nathanmarz:
Made an inline comment about using an existing function to clean up the code.

d2r:
Updated, thanks for the tip.

test/clj/backtype/storm/security/serialization/BlowfishTupleSerializer_test.clj
```diff
@@ -0,0 +1,63 @@
...
+        [backtype.storm.util :only (exception-cause?)]
+  )
+  (:import [backtype.storm.security.serialization BlowfishTupleSerializer]
+           [backtype.storm.utils ListDelegate]
+           [com.esotericsoftware.kryo Kryo]
+           [com.esotericsoftware.kryo.io Input Output]
+  )
+)
+
+(deftest test-constructor-throws-on-null-key
```
nathanmarz, line 13: This test isn't right. If it throws no exception, then nothing happens. You also don't need that let, you can just do {} directly in the constructor. You should do something like this:

`(is (thrown? RuntimeException (new BlowfishTupleSerializer nil {})))`

test/clj/backtype/storm/security/serialization/BlowfishTupleSerializer_test.clj
```diff
@@ -0,0 +1,63 @@
...
+      (new BlowfishTupleSerializer nil conf)
+      (catch Throwable t
+        (is (exception-cause? java.lang.RuntimeException t)
+        "Throws RuntimeException when no encryption key is given.")
+      )
+    )
+  )
+)
+
+(use '[clojure.string :only (join split)])
```
nathanmarz: Put the declaration in the ns form. `use` is for the REPL.

test/clj/backtype/storm/security/serialization/BlowfishTupleSerializer_test.clj
```diff
@@ -0,0 +1,63 @@
...
+        reader-bts (new BlowfishTupleSerializer kryo storm_conf)
+        buf-size 1024
+        output (new Output buf-size buf-size)
+        input (new Input buf-size)
+        strlist (split test-text #" ")
+        delegate (new ListDelegate)
+       ]
+    (-> delegate (.addAll strlist))
+    (-> writer-bts (.write kryo output delegate))
+    (-> input (.setBuffer (-> output (.getBuffer))))
```
nathanmarz: I think this is clearer like this:

`(.setBuffer input (.getBuffer output))`

nathanmarz:
Made a few more comments – looks like that first test doesn't test what you're trying to test.

d2r:
Fixed the test and cleaned up the code as per the comments.

nathanmarz:
I just tried running the test and this was printed:

798  [main] ERROR com.esotericsoftware.kryo.serializers.BlowfishSerializer - Encryption key not specified

Is this expected?

d2r:
It is, but I took a look, and that message is redundant given that we throw a RuntimeException anyway.  Also, the logger was pointing to kryo's BlowfishSerializer class.

Added a test for the invalid encryption key.

PR #478, opened 2013-02-14
Add support for building storm with maven

revans2:
This change adds support for building storm using maven in addition to leiningen.  It adds in some functionality that is nice for CI that leiningen does not currently support.
1. Test results in an xml format Jenkins understands
2. Code coverage using JaCoCo

The other files, in addition to m2-pom.xml, were changed to get them to compile with maven.  It seems to have something to do with the classpath and how clojure resolves class implementations that are written in clojure.  I have had the same issue with leiningen but not on all computers, so they may not be necessary.


nathanmarz:
I'd prefer to keep the build as Leiningen, as Leiningen as significantly easier to deal with (just look at the size difference between the build files). 

The Leiningen codebase is really easy to hack on. I bet it would be really easy to modify the existing `lein pom` task to make a plugin that generates the appropriate XML for Jenkins / JaCoCo.

revans2:
That seems fine. We will probably maintain a copy of the maven version until we can move over all of the same functionality to lein.

PR #497, opened 2013-03-09
Transition away from Java serialization for storing state on disk or in Zookeeper

hausdorff:
Addresses issue #419. Specifically, transitions `Supervisor` to clojure serialization, `StormTopology` to thrift, and maybe some other things I'm forgetting.

Also my first go-around with clojure. Any code written in clojure should be considered suspect.


hausdorff:
Is there something I can do to speed this patch integration?

jasonjckn:
Nathan, can you put backwards incompatible changes into 0.10 branch. 0.9.0-wip16 is super well tested and the only branch with metrics, rarely can you get this amount of testing , let's do compatible releases on it for while.

jasonjckn:
i'm going to be handling the merge of this fyi. i'll read through the code next chance I get.

jasonjckn:
(this will be going into 0.9.0 branch)

jasonjckn:
did you sign storm contributors agreement? if not please do.

jasonjckn:
Does this PR solve https://github.com/nathanmarz/storm/issues/525 e.g. can I add a new field to StormBase state for example, and the next time I start nimbus with this change all the previously created state that didn't have this field (but did have your PR) will still be readable. 

Is it forwards compatible as well?

hausdorff:
Great, looking forward to resolving the pr. I have signed the contributor's agreement.

It is "forwards compatible" in that serialized objects that are semantically identical are always unserialized to the same thing, regardless of whether we have done something like upgraded Java. It is less forwards compatible in that if you change the semantics of the storm state (and in particular LocalState) to demand something crazier than vanilla clojure serialization, you will need to build an accompanying implementation of the `backtype.storm.utils.StateSerializer` interface.

As for adding fields to StormBase, my response is limited to what I know about Storm (I'm a noob). But my answer is, I think that yes it should. Our proposed serialization scheme amounts to using standard clojure serialization, so for example the stormbase serialization here: https://github.com/nathanmarz/storm/pull/497/files#L0R304 should just work.

Also, if EDT is really a strict superset of clojure, then all serialization of LocalState should be compatible with anyone who wants to deserialize with implementation of EDT (I think this addresses questions you had in #525).

hausdorff:
What is the status of this PR?

ptgoetz:
I think it's safe to close this.

d2r:
This pull request was closed because it was superseded by #625.

PR #499, opened 2013-03-13
Build java classfiles with debugging symbols

mallman:
project.clj includes the line `:javac-options {:debug true}`, however this does not have the desired effect. Changing it to `:javac-options ["-g"]` works.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -43,7 +43,7 @@
   :repositories {"sonatype"
                  "http://oss.sonatype.org/content/groups/public/"}
 
-  :javac-options {:debug true}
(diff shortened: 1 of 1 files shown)


nathanmarz:
Can you send me a contributor agreement? It's annoying but unfortunately necessary to legally protect the project. Instructions are here, should only take a couple minutes: https://github.com/nathanmarz/storm/wiki/Contributing-to-Storm

mallman:
Sent.

Thanks for considering my patch. It makes debugging much easier.

Cheers,

Michael

PR #501, opened 2013-03-13
install_zmq.sh error handling logic

cfeduke:
install_zmq properly handles missing wget with curl (or missing curl with error), existing zeromq tar and dir, and git ignores downloaded artifacts


nathanmarz:
We're going to be getting rid of zmq soon, and this script is mostly just provided for documentation purposes, so I'm not going to accept any pull requests on this script.

PR #505, opened 2013-03-17
Wiped dead code that was shadowed by definition below

lazyval:
see also http://stackoverflow.com/q/15411876/298389


nathanmarz:
Thanks for pointing this out.

PR #514, opened 2013-03-21
Issue 260 add worker classpath config

b4hand:
This addresses Issue #260 with a slight addition of adding an option for adding to the nimbus classpath as well. This is useful for us as we have our own custom schedule which also has some HBase dependencies, so I found it pretty easy to add both at the same time.

I created this change off of the 0.8.3 branch because at the time, I couldn't get master to build via `lein`. The changes are small enough that the merge should apply cleanly regardless, but if you want me to rebase, let me know.


b4hand:
Looks like this brought in a bunch of merge commits that you probably don't want.

PR #517, opened 2013-03-22
Added SystemBolt.

jasonjckn:
SystemBolt has one executor per worker always. It's perfect for
exporting worker related metrics data.

Unit tests past 30 times in row, i've been testing this for a full day on a test storm cluster.


jasonjckn:
I had to change unit tests a bit, because the original test as is hangs at assert-failed nathanmarz#512 I think there is an issue with simulated time & message timeouts specifically that independent of this diff, and metrics tests. I spent 6 hours trying to figure it out, I tried rewriting simulated time in my own version, and waiting on all the disruptor queues to be empty during simulated time steps, but that didn't fix it.

But the new unit tests still tests the -exact- same stuff, just the timeout numbers are slightly different and it seems to work.
I ran these unit tests 60 times in a row and it passed each time.

src/clj/backtype/storm/daemon/supervisor.clj
```diff
@@ -85,8 +85,8 @@
   (let [local-assignment (assigned-executors (:port worker-heartbeat))]
     (and local-assignment
          (= (:storm-id worker-heartbeat) (:storm-id local-assignment))
-         (= (set (:executors worker-heartbeat)) (set (:executors local-assignment))))
-    ))
+         (= (disj (set (:executors worker-heartbeat)) Constants/SYSTEM_EXECUTOR_ID)
```
nathanmarz, line 88: I wonder if there's a better way to do this. This feels a bit dirty.

src/clj/backtype/storm/daemon/supervisor.clj
```diff
@@ -117,7 +117,12 @@
 (defn- wait-for-worker-launch [conf id start-time]
   (let [state (worker-state conf id)]    
     (loop []
-      (let [hb (.get state LS-WORKER-HEARTBEAT)]
+      (let [hb (try (.get state LS-WORKER-HEARTBEAT)
```
nathanmarz: This should be a separate patch.

src/jvm/backtype/storm/metric/SystemBolt.java
```diff
@@ -0,0 +1,164 @@
...
+                if (doEvent) {
+                    doEvent = false;
+                    return 1;
+                } else return 0;
+            }
+        }, bucketSize);
+
+        // This is metric data global to topology, but we don't support this concept, so put it here.
+        // It's very useful to have time series of TOPOLOGY_WORKERS to compare actual worker count.
+        context.registerMetric("configTopologyWorkers", new IMetric() {
```
nathanmarz: This is the originally configured worker count, not the current worker count. This is not particularly useful.

nathanmarz: Though I do like the idea of having Nimbus emit stats streams for global information. That could be a separate feature/pull request.

src/jvm/backtype/storm/metric/SystemBolt.java
```diff
@@ -0,0 +1,164 @@
...
+            }
+        }, bucketSize);
+
+        // You can calculate topology percent uptime between T_0 to T_1 using this metric data:
+        //       let s = sum topologyPartialUptimeSecs for each worker for each time buckets between T_0 and T_1
+        //       topology percent uptime = s/(T_1-T_0)
+        // Even if the number of workers change over time the value is still correct because I divide by TOPOLOGY_WORKERS.
+        context.registerMetric("topologyPartialUptimeSecs", new IMetric() {
+            private long _prevUptime = jvmRT.getUptime();
+            private final double NUM_WORKERS = RT.doubleCast(stormConf.get(Config.TOPOLOGY_WORKERS));
```
nathanmarz: TOPOLOGY_WORKERS != current number of workers, so this is wrong. The metrics here should be kept to being worker-specific.

nathanmarz:
Overall, this is good. Some comments:
1. Let's split out changes in here not related to system bolt to separate patches, as noted in the inline comments.
2. Let's get rid of the global stats and keep the system bolt to tracking worker-specific stats. I think it's a good idea to have a stats stream coming from Nimbus for global stats info.

src/clj/backtype/storm/daemon/supervisor.clj
jasonjckn, line 88: I could disj it at the worker.clj level, so that system executor doesn't heart beat. 
The semantics of this supervisor function are still consistent 
(defn matches-an-assignment? [worker-heartbeat assigned-executors]
Yes you need to disj to do matching check because scheduler never assigns a system executor. And it shouldn't.
Can you elaborate? I thought a lot about this, and this is what I came up with.

src/clj/backtype/storm/daemon/supervisor.clj
jasonjckn: I'll move it to a new pull request... but when you merge a pull request all the commits get squished anyways... it's effectively like just moving this it's own commit.

jasonjckn: Here you go: https://github.com/nathanmarz/storm/pull/523

src/jvm/backtype/storm/metric/SystemBolt.java
jasonjckn: I understand that, that's the purpose of this metric, the uptime with respect to requested number of workers. This really useful.
If someone doesn't get all the workers they request OR if they're having particular workers failing, the computed topology uptime will go below 100% which is what you want. If you strictly just count the current number of workers right now, you'll get uptime of 100% even when workers are dying. I think what you're hinting at is you could look at the scheduler to see situations like with the even scheduler where the scheduler chooses to allocate less workers then requested and this could be considered 100% uptime, I don't count that as 100% uptime and it's incredibly cumbersome to get the number from the scheduler. This stat is awesome, because even as people change their setNumWorkers over time, the topologyUptime metric will still feed you a value between 0% and 100% which factors in the timeline of different setNumWorker requests. 

please confirm you want this removed.

src/jvm/backtype/storm/metric/SystemBolt.java
jasonjckn: nimbus level and topology level information is really useful, but let's refactor this code in the future that's big PR, i only needed topology level info once here and this solves it. 

please confirm you want me to remove.

jasonjckn:
Commented & Pushed new changes.

jasonjckn:
one new PR #523

src/jvm/backtype/storm/metric/SystemBolt.java
jasonjckn: i'll remove this, i'm going to fork storm for twitter, so that I can add stuff like this back-in because I need it now. But clearly this would be better as a nimbus stat which factors in scheduling decisions.

src/clj/backtype/storm/daemon/supervisor.clj
nathanmarz, line 88: Yea, let's just leave it as is.

src/jvm/backtype/storm/metric/SystemBolt.java
```diff
@@ -0,0 +1,139 @@
...
+        GarbageCollectorMXBean _gcBean;
+        Long _collectionCount;
+        Long _collectionTime;
+        public GarbageCollectorMetric(GarbageCollectorMXBean gcBean) {
+            _gcBean = gcBean;
+        }
+        @Override
+        public Object getValueAndReset() {
+            Long collectionCountP = _gcBean.getCollectionCount();
+            Long collectionTimeP = _gcBean.getCollectionCount();
```
ankitoshniwal, line 60: Hey Jason,

Should it be _gcBean.getCollectionTime(); ?

Thanks,
Ankit.

PR #519, opened 2013-03-25
0.9.0 cluster state inimbus

jasonjckn:
This pull gives me access to cluster-state in INimbus, which is needed for this pull https://github.com/nathanmarz/storm-mesos/pull/5

"IMPORTANT: CuratorFramework instances are fully thread-safe. You should share one CuratorFramework per ZooKeeper cluster in your application." https://github.com/Netflix/curator/wiki/Framework

MesosNimbus needed zk style state, so this pull gets it done. Without making a 2nd curator framework.

I had to use Object for clusterState because ClusterState is a protocol, and therefore not accessible due to compilation ordering where javac goes first. If someone needs ClusterState in java in the future, we can rewrite it without protocols, but there's no sense changing all that code right now when only mesosnimbus needs it, and I wrote it partly in clojure.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(defproject storm/storm "0.9.0-wip15"
+(defproject storm/storm "0.9.0-wip16"
   :url "http://storm-project.clj"
   :description "Distributed and fault-tolerant realtime computation"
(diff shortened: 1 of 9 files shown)


jasonjckn:
i did this for 0.9.0 branch because it's an api change.

PR #520, opened 2013-03-25
pass cluster-state to INimbus

jasonjckn:
This pull gives me access to cluster-state in INimbus, which is needed for this pull https://github.com/nathanmarz/storm-mesos/issues/5

"IMPORTANT: CuratorFramework instances are fully thread-safe. You should share one CuratorFramework per ZooKeeper cluster in your application." https://github.com/Netflix/curator/wiki/Framework

MesosNimbus needed zk style state, so this pull gets it done. Without making a 2nd curator framework.

I had to use Object for clusterState because ClusterState is a protocol, and therefore not accessible due to compilation ordering where javac goes first in storm source code.  The clojure compiler does generate a class byte code for ClusterState but that comes after javac is finished. So other projects which depend on storm can merely cast the argument clusterState to ClusterState.

### DIFF
--- a/src/clj/backtype/storm/daemon/nimbus.clj
+++ b/src/clj/backtype/storm/daemon/nimbus.clj
@@ -42,12 +42,13 @@
     scheduler
     ))
 
-(defn nimbus-data [conf inimbus]
+(defn nimbus-data [conf inimbus cluster-state]
   (let [forced-scheduler (.getForcedScheduler inimbus)]
     {:conf conf
      :inimbus inimbus
      :submitted-count (atom 0)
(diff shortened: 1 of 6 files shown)


nathanmarz:
I don't like the design of this. You shouldn't be exposing the cluster state directly like that. Besides just being dirty from an interface standpoint (random Object which is unclear how to use it), there's way too much opportunity here for abuse and subtle bugs. I would instead just add the ability to INimbus to return metadata for a newly created topology. And then that metadata can be stored in the topology base and provided in the TopologyDetails for the scheduler/INimbus implementation to use later. The metadata should be serialized as JSON. Gonna close this pull request in anticipation of a new one using a better design.

PR #523, opened 2013-03-26
This solves race condition in unit tests if you try to shutdown a worker...

jasonjckn:
... which cleans up worker state while you also try to wait for worker to launch by reading the same

state.

### DIFF
--- a/src/clj/backtype/storm/daemon/supervisor.clj
+++ b/src/clj/backtype/storm/daemon/supervisor.clj
@@ -112,7 +112,12 @@
 (defn- wait-for-worker-launch [conf id start-time]
   (let [state (worker-state conf id)]    
     (loop []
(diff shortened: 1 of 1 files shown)


nathanmarz:
I don't understand this patch. I recently changed LocalState to check for FileNotFoundException's. Also the state's for workers are independent as each worker is given a unique id on launching.

jasonjckn:
Which branch did you "changed LocalState to check for FileNotFoundException" I saw nothing in master. 
The race condition is between https://github.com/nathanmarz/storm/blob/master/src/clj/backtype/storm/daemon/supervisor.clj#L141 and https://github.com/nathanmarz/storm/blob/master/src/clj/backtype/storm/daemon/supervisor.clj#L115

if L141 happens first then you get the exception.

PR #545, opened 2013-04-16
Millisecond-level tick generation granularity

tscurtu:
Issue #544.

Introduces a new timer scheduler for tick generation.

The nimbus and supervisor scheduled events are unaffected.


nathanmarz:
Closing this in the interest of using a backwards compatible design. See my comment on #544

PR #552, opened 2013-04-24
change default ackers number into the initial number of workers

lyogavin:
#377

1 change default config of ackers.executors into null
2 in the code, check when it's null, use the initial number of workers
3. change has-ackers functions to reflect the new change


mrflip:
+1

Is it just as good to say `(defn has-ackers? [storm-conf] (not= (storm-conf TOPOLOGY-ACKER-EXECUTORS) 0))` on [common.clj line 292](https://github.com/nathanmarz/storm/pull/552/files#L1R292)?

PR #555, opened 2013-04-30
Limit ShellBolt internal _pendingWrites queue length

xiaokang:
for storm issuse #541

### DIFF
--- a/storm-core/src/jvm/backtype/storm/task/ShellBolt.java
+++ b/storm-core/src/jvm/backtype/storm/task/ShellBolt.java
@@ -1,5 +1,6 @@
 package backtype.storm.task;
 
+import backtype.storm.Config;
 import backtype.storm.generated.ShellComponent;
 import backtype.storm.tuple.MessageId;
 import backtype.storm.tuple.Tuple;
@@ -55,7 +56,7 @@ public class ShellBolt implements IBolt {
     private ShellProcess _process;
     private volatile boolean _running = true;
     private volatile Throwable _exception;
-    private LinkedBlockingQueue _pendingWrites = new LinkedBlockingQueue(100);
+    private LinkedBlockingQueue _pendingWrites = new LinkedBlockingQueue();
     private Random _rand;
     
     private Thread _readerThread;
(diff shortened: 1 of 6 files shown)


storm-core/src/jvm/backtype/storm/task/ShellBolt.java
```diff
@@ -55,7 +55,7 @@
     private ShellProcess _process;
     private volatile boolean _running = true;
     private volatile Throwable _exception;
-    private LinkedBlockingQueue _pendingWrites = new LinkedBlockingQueue();
+    private LinkedBlockingQueue _pendingWrites = new LinkedBlockingQueue(100);
```
nathanmarz: This shouldn't be initialized here if its going to be initialized in the constructor.

storm-core/src/jvm/backtype/storm/task/ShellBolt.java
```diff
@@ -66,7 +66,12 @@ public ShellBolt(ShellComponent component) {
     }
 
     public ShellBolt(String... command) {
+        this(100, command);
+    }
+    
+    public ShellBolt(int maxQueueSize, String... command) {
         _command = command;
+        _pendingWrites = new LinkedBlockingQueue(maxQueueSize);
```
nathanmarz: This should be moved to prepare and be made configurable via the Storm config, not via constructor params.

nathanmarz:
Made some comments about changes that should be made. These are important because from a pure multilang definition, ShellBolt isn't constructed by the user so they need to be able to configure it via config.

xiaokang:
hi @nathanmarz , thanks for your comments. I adjust the code to config max pending in ShellBolt's prepare.

nathanmarz:
Looks like this needs to be rebased against master, as it can't be merged right now.

xiaokang:
Just rebased the branch against master but other's commits appear in this PR. Is it ok?

nathanmarz:
Hmm, that's really strange. That's definitely not ok. Maybe you can just close this one and open a new one against master?

xiaokang:
OK, #730 sent. Close this one.

PR #556, opened 2013-04-30
Netty based implementation for storm messaging

anfeng:
This pull request implements Storm messaging layer in Netty. 
- Client and server could be started in any order
- Messages will be delivered in batch if available
- Large messages are supported

The basic functionality has been verified via a simple topology test. I have also included some unit test to show its functionality.


nathanmarz:
Can you summarize how this works at a high level? I want to know:
1. How do initial connections work given that servers and clients start up asynchronously and the client may connect before the server is up?
2. What happens on failure to send a message?
3. Why is message sending happening asynchronously in this implementation?
4. How are connections closed / resources cleaned up when a worker moves or there's a shut down?
5. Why the changes to TaskMessage?

Also, let's move this implementation into its own module (Storm has a multimodule setup now). The Config.java changes can stay where they are, but everything else should go into a storm-netty module. To do this, add "storm-netty" to the MODULES file, and then base the build for storm-netty off of the storm-core project.clj.

anfeng:
When client starts, we establish a local  queue, and then try to connect to the specified server. The message queue enables app layers to invoke send(task, message) even if the client has not established its connection. The connect() operation will call 
    (1) StormClientHandler::channelConnected() if connection was estabished 
or (2) StormClientHandler::exceptionCaught() if connection attempt failed

When  StormClientHandler::channelConnected() is invoked, we take messages (if any) from the associated queue and send them to server.  If  StormClientHandler::exceptionCaught()  is invoked, our remote server is not ready. We use an exponential backoff policy to wait a little, and then retry connection (up to max # of tries).

Server side logic is simpler but similar. When a server starts, we create a queue so that messages could be pulled by app layer, and tries bind() against a local port. Upon a connection from a client, a Netty Channel is established and a StormServerHandler is created to deal with that connection. Besides bookkeeping channels, StormServerHandler mainly deals with incoming message, and put them into the associated queue. Then, server sends back an ACK response to client. Upon receiving an ACK response, client will try to deliver remaining messages (if any).

Client sends messages to server via channel.write() in StormClientHandler.java. If that operation failed, we  close the current socket. The current implementation will drop all those messages that we just failed to send. 
- Question: should we resend those messages before closing socket?

Current implementation uses "async" extensively for a few reasons:  
- Netty is an asynchronous framework, and it achieve high throughput because of that design. 
- Storm needs clients to be able to launch even before server exist, and thus we introduced a "queue" b/w 
  message publisher and underlying delivery layer. This queue based delivery enables us to deliver messages 
  in batches. 

Context.term() should be called when a worker is shutdown. Context.term() is implemented to invoke close() method of all  Client and Server objects associated with this Context. 
- When Client.close() is called, we put a "CLOSE" message into its queue. Client delivery thread will send off existing messages, and then invoke Client.close_n_release() method which actually close connection and release resources. 
- When Server.close() is called, we simply close all associated channels (server channels and client channels), and release resources.

TaskMessage.java was modified to support message with NULL payload. Maybe I should use a separated pull request. Should I?

Per your suggestion, I moved Netty implementation into storm-netty. I'm not sure how its project.clj could state its dependency on storm-core, and thus have hard-coded project.clj now. Please offer advise.

anfeng:
project.clj has been revised to follow other modules. storm-netty added into MODULES.

nathanmarz:
Merged! I removed the benchmark since that doesn't belong in the unit tests. Can you submit that to @stormprocessor/storm-benchmark?

nathanmarz:
Actually one question – is the behavior now to drop on failure or to retry and potentially deliver a message twice? It should be the former to meet Storm's requirements. So the delivery model should be at most once and NOT at least once.

anfeng:
at most once.

anfeng:
Very cool. We are one step close to Apache.

For the benchmark code, do you take clj code? Your existing code is in Java.

nathanmarz:
Yes, that repo is just intended to just be a collection of all the benchmark stuff we develop for Storm. Let's just add a clj folder to put this benchmark.

tolitius:
@anfeng What is the rationale behind using "LinkedBlockingQueue" for internal queues vs. an off heap storage (e.g. a direct byte buffer)? Wouldn't "LinkedBlockingQueue" pose GC, Heap and Blocking problems that are currently solved by 0mq?

PR #563, opened 2013-05-03
Enable null payload in TaskMessage 

anfeng:


nathanmarz:
I don't understand the purpose of this. When does Storm need to send null payloads?

anfeng:
For netty implementation, we need some system-level messages (see https://github.com/anfeng/storm/blob/messaging/storm-netty/src/jvm/backtype/storm/messaging/netty/Util.java):
- End-of-batch message ... indicate as the last message of a request batch
- Response message (OK/FAILURE) ... inform sender that requests have been received

To simplify our implementation, I like to use TaskMessage for these system messages. 
These system messages don't need payload, and just need to use "taskID". 
This is similar to what we do today to inform receiving thread to shutdown by  
a TaskMessage with taskID=-1 in 
https://github.com/anfeng/storm/blob/messaging/storm-core/src/clj/backtype/storm/messaging/loader.clj

nathanmarz:
It seems to me that this is an abuse of TaskMessage. The taskId=-1 thing was a hack, I don't like that idea of continuing to abuse it. It seems to me that this sort of stuff that you need to do should be done by the transport mechanism, completely isolated from TaskMessage.

anfeng:
Agree. Let's close this pull request. I have introduced ControlMessage within Netty implementation instead. See the revised pull request https://github.com/nathanmarz/storm/pull/556

PR #566, opened 2013-05-09
Ignore NoNodeExists exceptions

jasonjckn:


src/jvm/storm/trident/topology/state/RotatingTransactionalState.java
```diff
@@ -96,7 +98,16 @@ public void cleanupBefore(long txid) {
         SortedMap<Long, Object> toDelete = _curr.headMap(txid);
         for(long tx: new HashSet<Long>(toDelete.keySet())) {
             _curr.remove(tx);
-            _state.delete(txPath(tx));
+            try {
+                _state.delete(txPath(tx));
+            } catch(RuntimeException e) {
+                // Ignore NoNodeExists exceptions because when sync() it may populate _curr with stale data since
+                // zookeeper reads are eventually consistent.
+                if(!(e.getCause() instanceof KeeperException.NoNodeException)) {
```
nathanmarz: Let's replace this with logic similar to the exception-cause? function used in Clojure. That will be more robust than just checking one level deep in the exception stack.

nathanmarz:
Just had a single comment about improving the code.

jasonjckn:
Do you ever call clojure from java? Now we have code duplication with the above commit. 

If you'd like I can have clojure invoke the above java function.

nathanmarz:
Yea, we should just have the clojure code call the Java version.

nathanmarz:
Oh, I just realized this was against 0.9.0 branch. Make pull requests against master, not 0.9.0.

jasonjckn:
ok, next time, please cherry-pick it.

PR #567, opened 2013-05-09
Update project.clj - Bump snakeyaml dependency version from 1.9 to 1.11 (Latest Release)

kb:
Storm is using a snakeyaml version that is now two revisions old. I have recently come across a dependency conflict when using Storm with JRuby 1.7.3. When running in ruby 1.9 mode JRuby utilizes the snakeyaml library, however the internal JRuby  version of snakeyaml 1.11. Since Storm is running an outdated version, it would make sense to propose the change in hopes of resolving this dependency conflict.

### DIFF
--- a/storm-core/project.clj
+++ b/storm-core/project.clj
@@ -19,7 +19,7 @@
                  [org.clojure/tools.logging "0.2.3"]
                  [org.clojure/math.numeric-tower "0.0.1"]
                  [storm/carbonite "1.5.0"]
-                 [org.yaml/snakeyaml "1.9"]
+                 [org.yaml/snakeyaml "1.11"]
                  [org.apache.httpcomponents/httpclient "4.1.1"]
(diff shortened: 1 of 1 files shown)


colinsurprenant:
+1 

JRuby 1.7.3, which will be included in the upcoming RedStorm release, uses snakeyaml 1.11. 

We cannot downgrade snakeyaml to 1.9 in the topology jar - when running a topology that uses the JRuby YAML api, it crashes on org.yaml.snakeyaml.events.DocumentStartEvent.getVersion() which returns an Integer object in 1.9 and a DumperOptions$Version object in 1.11. 

Upgrading snakeyaml to 1.11 in Storm 0.8.2 has solved the problem and basic/test topologies are working well. I haven't done any further analysis & tests in Storm to verify potential problems with this upgrade.

PR #576, opened 2013-05-28
fixes memory leak in trident.

jasonjckn:
LRUMap was being accessed concurrently. This patch prevents the
concurrent access of LRUMap (backed by non-concurrent data structure).


jasonjckn:
Let me know if you want this in a different branch than 0.9.0.

nathanmarz:
Thanks for tracking this down. I ended up doing this instead: https://github.com/nathanmarz/storm/commit/d96f4313598069c54d26b8627f9bf65dfe2f3fa7

PR #582, opened 2013-06-03
Try at config validation

d2r:
For issue #581

This pull request is for input to see if this is a direction we could go.

In thinking about how to fix this sort of issue, I figure we want to validate the config at the point of topology submission, since after that it enters ZK state and would become a nuisance.

The idea is to use the defaults.yaml file we already have to validate types coming in with the topology.

Pros:
- Should only have to change the defaults.yaml to change the schema.
- Can be generalized to lists and maps

Cons:
- Need defaults for every config where the type truly matters.
  - Would have to change storm code to deal with null (e.g. use the empty String instead)
  - What if there are no sensible defaults?
- Lists with differing types prescribed by elements would not be enforced.

The code in this pull is incomplete, but I'm looking for comments on the direction.

### DIFF
--- a/storm-core/src/dev/test-valid-non-collection-types.yaml
+++ b/storm-core/src/dev/test-valid-non-collection-types.yaml
@@ -1,0 +1,1 @@
+topology.workers: 3
--- a/storm-core/src/jvm/backtype/storm/nimbus/BasicTopologyValidator.java
+++ b/storm-core/src/jvm/backtype/storm/nimbus/BasicTopologyValidator.java
@@ -23,7 +23,7 @@ public class BasicTopologyValidator implements ITopologyValidator {
 
         while (it.hasNext()) {
             while (! toRet.isInstance(it.next())) {
-                toRet = toRet.getSuperClass();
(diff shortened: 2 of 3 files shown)


nathanmarz:
I don't like inferring types from the defaults.yaml for the reasons you mentioned. Let's put the schema information in the Config.java. For backwards compability, we can't change the values of those elements to have more information, so I suggest for every field we add schema information by making a corresponding field with "SCHEMA" appended to it. e.g. TOPOLOGY_WORKERS_SCHEMA. Then we can modify config.clj to verify that every field have a corresponding schema, and the config.clj can generate a map from config name -> schema function.

I think the best thing to do for the schema fields is to make them polymorphic. For simple fields we can just put something like Number.class or String.class, but for more complex fields we may want to code up an arbitrary function. So the schema fields should be of type Object. A schema field set to Number.class would be an implicit function like (fn [field](instanceof Number field)). To make things appropriately Java-like we can make that function be an interface, such as:

``` java
interface FieldValidator {
  boolean validateField(Object field);
}
```

d2r:
Opened #583

PR #583, opened 2013-06-04
Add config schema validation

d2r:
For issue #581

This pull request supersedes #582

This is a prototype implementation to continue along the path suggested in this comment https://github.com/nathanmarz/storm/pull/582#issuecomment-18880232

Adds a schema to each config, and validates on reading the config yaml.

Just looking for comments and direction at this point.

TODO:
- There should be a single validation function: `validate-configs-with-schemas` should just call this function and not have to test for `nil` or `instanceof FieldValidator`.  The mapping created in `b.s.config` should map the field value to this function instead of to the schema.
- We should validate on `findAndReadConfigFile`, and not on `readStormConfig`, since it will unnecessarily validate the defaults.yaml.  (It is currently there for testing.)
- Some of the types are not sufficient.  Some cases of `java.lang.Number` might need to be `Integer` instead
- Some reorganization of code (Do we want `FieldValidator`s in `Config`?)
- Lots of code intended for debugging needs to be removed.

### DIFF
--- a/storm-core/src/clj/backtype/storm/config.clj
+++ b/storm-core/src/clj/backtype/storm/config.clj
@@ -1,6 +1,6 @@
 (ns backtype.storm.config
   (:import [java.io FileReader File])
-  (:import [backtype.storm Config])
+  (:import [backtype.storm Config Config$FieldValidator])
   (:import [backtype.storm.utils Utils LocalState])
   (:import [org.apache.commons.io FileUtils])
   (:require [clojure [string :as str]])
@@ -24,6 +24,24 @@
   (dofor [f (seq (.getFields Config))]
          (.get f nil)
          ))
+
+;; Create a mapping of config-string -> validator
+;; Config fields must have a _SCHEMA field defined
+(def CONFIG-SCHEMA-MAP
+  (reduce (fn [mp [k v]] (assoc mp k v)) {}
+    (for [f (seq (.getFields Config))
+      :when (not (re-matches #".*_SCHEMA$" (.getName f)))
+      :let [k (.get f nil)
+            nam (.getName f) 
+            field-to-get (str/replace nam #"$" "_SCHEMA")
+;            _ (println (str "Trying to set " k " to " field-to-get " value"))
+            v (-> Config (.getField field-to-get) (.get nil))
+;            _ (println (str "and value was " v ))
+            ]]
+      [k v]
+    )
+  )
+)
 
 (defn cluster-mode [conf & args]
   (keyword (conf STORM-CLUSTER-MODE)))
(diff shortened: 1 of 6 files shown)


storm-core/src/clj/backtype/storm/config.clj
```diff
@@ -62,11 +80,40 @@
 (defn read-default-config []
   (clojurify-structure (Utils/readDefaultConfig)))
 
+(defn- validate-configs-with-schemas [conf]
+  (doseq [[k v] conf
+         :let [_ (println (str "trying to validate" [k v]))
```
nathanmarz: These printlns should be removed

storm-core/src/clj/backtype/storm/config.clj
```diff
@@ -62,11 +80,40 @@
...
+               _ (println (str "schema is (" (get CONFIG-SCHEMA-MAP k) ")"))
+               schema (get CONFIG-SCHEMA-MAP k)]]
+    (if (instance? Config$FieldValidator schema)
+      (if (and (not (nil? v))
+               (not (.validateField schema v)))
+        (throw (IllegalArgumentException.
+                 (str "'" k "' " (.getCriteriaPredicate schema))))
+        (println (str "OK: " k " " (.getCriteriaPredicate schema)))
```
nathanmarz: This println should be removed

storm-core/src/clj/backtype/storm/config.clj
```diff
@@ -25,6 +25,24 @@
          (.get f nil)
          ))
 
+;; Create a mapping of config-string -> validator
+;; Config fields must have a _SCHEMA field defined
+(def CONFIG-SCHEMA-MAP
+  (reduce (fn [mp [k v]] (assoc mp k v)) {}
```
nathanmarz: More elegant to do something like:

``` clojure
(->> (.getFields Config)
       (filter #(not (re-matches #".*_SCHEMA$" (.getName %))))
       (map (fn [s] [s (get-field-validator (.getField Config (str s "_SCHEMA"))])
       (into {}))
```

And get-field-validator can be a multi-method that transforms classes into an appropriate field validator, throws an error on nil, and if the value is already a field validator returns it.

```
```

nathanmarz: This also allows you to greatly simplify the validation code below.

nathanmarz:
Made some inline comments. Please take note especially of the one about restructuring the code to greatly simplify it and make it more extensible at the same time to boot. Let's also get rid of all the printlns.

d2r:
Agreed.  Will do.  Thank you.

nathanmarz:
Any update on this? Would really like to get this merged in.

d2r:
I'll take a look again tonight.

d2r:
- Removed debug code
- Restructured validation function as per @nathanmarz suggestions
- Moved validators into a new file so that I could write tests.  Calling getFields on Config was collecting the validators along with the actual config fields & schemas.
- Tests do not cover every config & schema.  I wrote one test to check the validation of TOPOLOGY_WORKERS, since that is what caused #581.
- I did not change when configs were validated, because I did not see a good way to avoid validating the defaults.  One effect is that validation code will run at compile time.
- Regarding Number/Integer, it seems that Number is being coerced to int when we call from clj->java method.
- Fixed a new default that was not added to Config.java
- Unit tests passed.

nathanmarz:
Awesome work. Merged.

storm-core/src/jvm/backtype/storm/ConfigValidation.java
```diff
@@ -0,0 +1,82 @@
...
+            if (o == null) {
+                // A null value is acceptable.
+                return;
+            }
+            final long i;
+            if (o instanceof Number &&
+                    (i = ((Number)o).longValue()) == ((Number)o).doubleValue())
+            {
+                // Test whether the integer is a power of 2.
+                if (i > 0 && (i & (i-1)) == 0) {
```
xumingming, line 75: very interesting(good) method of validating whether a number is a power of 2!

PR #585, opened 2013-06-06
Specify jvm_opts when calling storm jar

srmelody:
This is a simple change to allow calling bin/storm jar with options that are passed from the command line.  We have a local change for this that we're using in house to deploy topology's to jenkins and having this in a release of storm would be great.


nathanmarz:
I don't have a contributor agreement from you. Can you send me one? Instructions here, should only take a couple minutes: https://github.com/nathanmarz/storm/wiki/Contributing-to-Storm

srmelody:
I sent you a signed contributor agreement.  Let me know if anything else is pending. 
Thanks!

PR #592, opened 2013-06-14
Add support for ShellBolt and ShellProcess objects to accept custom environment variables.

rbuck23:
We have a custom environment that needs this support.  This was brought up a couple of years ago.  It seems that writing a bash script wrapper is overhead that storm would want to avoid.

https://github.com/nathanmarz/storm/issues/32

### DIFF
--- a/storm-core/src/jvm/backtype/storm/task/ShellBolt.java
+++ b/storm-core/src/jvm/backtype/storm/task/ShellBolt.java
@@ -8,6 +8,7 @@ import backtype.storm.utils.ShellProcess;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.LinkedBlockingQueue;
 import static java.util.concurrent.TimeUnit.SECONDS;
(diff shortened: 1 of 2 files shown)


askrht:
Thanks for the contribution Randy. I reviewed your changes. Here are my suggestions / observations:
1. Wouldn't it be better to read the PATH of the executable from conf/defaults.yaml itself or add additional environmental variables to the same file? We could also add another yaml file to specify custom environment properties. Configuration files have an advantage over calling a java API to set the property simply because you can clearly see all the variables and avoid conflicts. Moreover, setting the environment variable dynamically will not have any affect after the ShellProcess has been created so it does not add any additional value unless you were passing a password or some other sensitive information. Even sensitive information can be stored in a file by encrypting the value. 
2. If we are going to support this for ShellBolt, we should support them for ShellSpout as well. Another benefit of using a config file is that it is centralized therefore we do not have to modify ShellBolt, ShellSpout or any other component that may be added in future.
3. It is a good idea to add the following unit tests or create a separate task for them.
   a) test the functionality existing public constructors that you have modified.
   b) test the functionality of new public constructors and methods added by you.

Conclusion:
I would suggest carefully analyzing point number one or discuss an alternative implementation because once you make your API public, there is no going back and you have to live with it forever.

PR #595, opened 2013-06-14
Better error logging and faster restarting of dying workers by the supervisor.

knusbaum:
Fix for: https://github.com/nathanmarz/storm/issues/489
(Launching a process throws away standard out; can hang)


nathanmarz:
Can you give a high level overview of what this implementation does? From glancing at it, I don't like that it launches a thread per launched worker and the use of global state in the dead worker functions. Any state should be kept in the supervisor-state map.

knusbaum:
When a worker process is spawned, a thread is created to log its standard output and then wait for its termination. On termination, the process is marked as dead in the dead-workers structure. 

If the worker isn't supposed to be dead, read-allocated-workers will notice that one of its workers is in the dead-workers structure, log that fact, and immediately attempt to restart the process.

I couldn't find a nice clean way around spawning a thread for each worker process, and since these threads are going to be blocked for the vast majority of their lives, I thought it was acceptable. I did look into making the processes into one thread, but two problems needed to be overcome:

1: Merging input streams so that a single thread can do a blocking read on all workers' outputs
2: Waiting on multiple processes to die on a single thread. C has wait(), but I was surprised not to be able to find anything similar in Java. Any of the abstractions proposed on threads I read online ultimately used threads to wait on individual processes. The only solution I came up with for this was to busy-wait on the processes, using a timer to periodically interrupt the wait and iterate the process list.

The threaded solution seemed much cleaner for the small overhead involved.

The dead-workers data structure, on the other hand, I will move into the supervisor-state map.

nathanmarz:
I really don't like the threaded solution. Threads are a resource that if they fail to be properly cleaned up can cause weird issues to the supervisor. I think your proposal #2 is the way to go.

The way it's implemented now the logging will go to the supervisor log file, which definitely is not what we want. Can we get the logging to go to a separate worker log file for each worker? I think this will require controlling the classname the log messages are under and then an update to the logback config. 

The dead-workers data structure seems to be an optimization that's independent from redirecting the worker output. Can we remove all of that from here and re-open it as its own pull request once we finish this?

knusbaum:
I've removed the dead-workers optimization entirely and am working on the rest, although instead of a separate thread as I initially suggested, I'm currently checking the workers' output/status in supervisor/sync-processes (an option I overlooked).
It can be easily moved into its own thread if you want, but it doesn't seem necessary to me.

One issue: 
Without threads, I need a bit of state to keep track of the Process objects in order to check the exit code, but Process objects are not serializable.
Since the objects are meaningless to another instance of the supervisor anyway, it seems reasonable to track the objects outside the supervisor-state. What do you think?

knusbaum:
Working on logging into the workers' log files, but it doesn't look like a simple change of the logback config along with controlling the classname. I don't see any way of dynamically choosing a log file except with JoranConfigurator, which is a bit ugly.

storm-core/src/clj/backtype/storm/util.clj
```diff
@@ -417,6 +419,22 @@
...
 
+;; Logs stdout of a Java Process, and monitors of 
+(defnk mk-watch-fn [process :log-fn nil :exit-fn nil]
+  (let [reader (BufferedReader.
+                (InputStreamReader.
+                 (.getInputStream process)))]
+    (fn []
+      (if log-fn
+        (while (.ready reader)
+          (log-fn (.readLine reader))))
```
knusbaum: I'll fix this. It's not non-blocking.

knusbaum:
I'll argue against trying to get the output of dying worker processes into the individual worker logs for two reasons:

1: Standard output from workers tends to occur when the JVM fails to launch due to improper supervisor configuration, not when there's something wrong with the worker.
2: This is a rare occurrence, and there doesn't appear to be an idiomatic way to report these errors to the workers' logs.

nathanmarz:
Standard output could also occur because someone was doing some debugging locally and accidentally deploys the topology to the cluster with stdout code. With logback we can control the size of the log files and prevent topologies like this from sucking up all the disk space on the cluster (and thereby causing a cascading failure). This is an important case to cover and not rare enough to ignore.

d2r:
It's possible to hook the worker's stdout/stderr to its logger, once the logger has been initialized.

Before the workers logger is initialized, it's unclear to me how to accomplish this.

If the supervisor simply opens and appends to the appropriate worker log directly, then it needs to be aware of the log file's path.  That means, I think, the supervisor needs to read the logging configuration and evaluate the template plugging in the correct port.  This also has the drawback that we miss other appenders configured, e.g., to send worker logs to a remote machine.

Or, if the supervisor initializes another logger for the worker that is launching, then it seems we could not use the same logging configuration, as setting the port in the logging system would necessarily affect the logging system across all worker launches.  I could be wrong about this, but I thought the logging context was static even across all threads.

Could there be a solution where we log the worker stdout/stderr to the supervisor log only until the worker initializes its own logger and can redirect its own stdout/stderr itself?

PR #597, opened 2013-06-18
build_release.sh  modified to avoid duplicated storm-core.jar files

anfeng:
Some Storm modules (ex. storm-netty) depend on storm-core. We need to modify build_release.sh so that storm-core-VERSION.jar is not packaged in storm.home and its lib subdir.


xiaokang:
Hi @anfeng, is it more general to remove all $DIR/*.jar from $DIR/lib/ ? Since other jars may be copied to lib if it's a dependency of other modules.

nathanmarz:
I agree with @xiaokang

anfeng:
OK, it's revised per @xiaokang

nathanmarz:
Shouldn't that be cd ../.. at the end?

anfeng:
Nathan, I have revised the "cd ../.."

PR #598, opened 2013-06-18
add web ui for :

xiaokang:
1. viewing log file
2. get/set log4j log level online

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@
  * Append component name to thread name of running executors so that logs are easier to read
  * Messaging system used for passing messages between workers is now pluggable (thanks anfeng)
  * Netty implementation of messaging (thanks anfeng)
+ * Include topology id, worker port, and worker id in properties for worker processes, useful for logging (thanks d2r)
+ * Tick tuples can now be scheduled using floating point seconds (thanks tscurtu)
  * Bug fix: Supervisor provides full path to workers to logging config rather than relative path (thanks revans2) 
  * Bug fix: Call ReducerAggregator#init properly when used within persistentAggregate (thanks lorcan)
  * Bug fix: Set component-specific configs correctly for Trident spouts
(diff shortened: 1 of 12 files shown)


xiaokang:
for issuse #543

nathanmarz:
What happens to the supervisor if something goes wrong with the internal log UI? What if the string being returned is 1GB in size, for example? We need to make sure that these edge conditions are accounted for, or that the supervisor crashes with an exception whenever they happen.

xiaokang:
Hi nathan, thanks for review.

If something goes wrong (such as log file does not exist) with the log UI, an exception message will be logged by Jetty.

The code limit the string returned to 10000 lines and 10MB by reading n (default 1000, max 10000) lines from the read point (file length - tail length(deault 10KB, max 10MB)) of the logfile. A very extreme edge condition that the log file grows very fast and a single line starting from the read point is larger than 10MB will cause large memory consumption.

nathanmarz:
Yea, let's remove the logic around number of lines and only read form the file in terms of number of bytes. That way we can just set an upper limit to how much can be returned and protect against any memory issues.

xiaokang:
OK, I will try to implement this and update the pull request.

nathanmarz:
Actually, on second thought – can we just make this its own independent daemon process? Isolating it from the supervisor process ensures something going wrong in the log viewer won't affect the supervisor (or is at least much less likely to).

xiaokang:
Isolating it from the supervisor process is indeed more safe for supervisor but it's more complicated to deploy a storm cluster. So it may be balance between easy to use and very rare errors. 

It should be safe to integrate it in supervisor since viewing web UI log is usually issued by by human not so frequently.I have made some change to the code to ensure that only read some byte (default 10KB, max 10MB) from tail of log file. Hope that it will be more acceptable since our users feed back that web UI for logs is really useful.

nathanmarz:
I thought about it more and this really should be done as its own daemon. It barely makes things more complicated, It's just another process that needs to be daemonized. In almost all setups this should be a one line deployment change. Isolation of functionality is always worth it when the cost of doing so is so low. Let's go ahead and add a log viewer daemon to backtype.storm.daemon and add the appropriate command to the bin/storm script to launch it.

xiaokang:
OK, I agree with that it's worth to doing so for isolation and system stability. I will update the code soon.

xiaokang:
Just updated the code.

PR #600, opened 2013-06-21
remove storm-*.jar from lib dir

xiaokang:
storm-0.9.0.wip19.jar is copied to lib dir as a dependency of other modules. So storm can not startup due to multiple conf file in classpath. Just delete duplicated top jar files in lib dir.


xiaokang:
Oh. the pull request is duplicated with #597. Close it.

PR #602, opened 2013-06-21
Launch workers with id and port properties

d2r:
Launch workers with id and port properties

When logging to a central location with a logging utility like syslog,
make it possible to split logs using logging configuration.

This greatly helps, e.g., when viewing logs when many different large
topologies running concurrently.

### DIFF
--- a/storm-core/src/clj/backtype/storm/daemon/supervisor.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/supervisor.clj
@@ -418,6 +418,9 @@
                        " -Dlogfile.name=" logfilename
                        " -Dstorm.home=" storm-home
                        " -Dlogback.configurationFile=" storm-home "/logback/cluster.xml"
+                       " -Dstorm.id=" storm-id
+                       " -Dworker.id=" worker-id
+                       " -Dworker.port=" port
                        " -cp " classpath " backtype.storm.daemon.worker "
                        (java.net.URLEncoder/encode storm-id) " " (:assignment-id supervisor)
                        " " port " " worker-id)]


kurtharriger:
I currently use the following to completely reconfigure logging on startup.  We use jenkins to automatically deploy new topologies after each commit so it is easier for me to re-configure the logging after startup then it is to modify the cluster wide logging configuration.  

I wish storm would allow you to specify a factory class and use that to create new bolts and spouts rather than serializing/deserializing them so that this logic could be centralized, but for now its just a synchronized static method.  I call this in every bolt on call to prepare and in every spout on call to open.  There are a few log messages emitted by storm itself to the worker log files all meaningful application messages occur after this and are redirected to graylog.

I embed the logging configuration for each environment in the jar file resources folder, e.g production.log4j.properties and specify this as a configuration property.  On application startup I specify the environment on the commandline and the application then loads property files corresponding to that environment adding/updating storm config properties including the resource file containing logging configuration. Note that the config object passed to open/prepare is not the same as what is passed to the submit so I also pass the full configuration object into the constructor of each bolt/spout.   Anyway sounds like you're trying to do something similar so I figured I'd share my solution if its useful to you.

```
import org.apache.log4j.LogManager
import org.apache.log4j.MDC
import org.apache.log4j.PropertyConfigurator

/**
 * Reconfigures log4j using log file specified in the configuration file
 *
 * The configuration file can be specified in topology configuration
 * and should be relative to the classpath root.
 * log4j.properties.file = <filename>
 *
 * Typically storm topologies share a logging configuration specified
 * on the storm cluster.  This makes it difficult to change logging levels
 * per topology for debugging or troubleshooting as the global configuration
 * must be changed on all supervisor nodes.  This class allows bundling
 * the log configuration with the topology and reconfiguring it when topology
 * is started.
 *
 * Storm unfortunately does not provide a place to put startup logic except
 * for the prepare method for each bolt and spout.  Therefore this logic
 * is synchronized and will only run once no matter how many times it has
 * been called.
 *
 * Existing loggers allocated with LogManager.getLogger will use the
 * new configuration after reconfigured and it is not necessary to
 * reallocate them (which is good since usually the logger is
 * allocated to static final field)
 *
 *
 */
class LogConfigurator {

    /**
     *
     * Reconfigures log4j using log file specified in the configuration file
     *
     * The configuration file can be specified in topology configuration
     * and should be relative to the classpath root.
     * log4j.properties.file = <filename>
     *
     * If no log4j.properties.file is specified in the configuration then logging
     * will not be reconfigured.
     *
     * Following the initial configuration MDC properties will be added from the
     * topology configuration and the system and configuration properties dumped
     * to the debug log.
     *
     * @param config The storm configuration map optionally containing log4j.properties.file
     *
     */
    public static void configure(Map config) {
        if(configured) return;

        synchronized (LogConfigurator.class) {
            if(configured) return;
            reconfigureLogging(config)
            addLogProperties(config)
            dumpConfigProperties(config)
            configured = true
        }
    }


    final static org.apache.log4j.Logger log = org.apache.log4j.LogManager.getLogger(LogConfigurator);
    private static volatile boolean configured = false;
    private static void reconfigureLogging(Map config) {
        String filename = config.get("log4j.properties.file")
        if (!filename)  return
        def stream = LogConfigurator.classLoader.getResourceAsStream(filename)
        if (!stream) {
            log.warn("Unable to find $filename on classpath")
        }

        Properties props = new Properties();
        props.load(stream)

        try {
            LogManager.resetConfiguration()
            PropertyConfigurator.configure(props);
            log.debug("Successfully reconfigured logging")
        }
        catch (Exception ex) {
            log.error("Failed to reconfigure logging", ex)
        }
    }

    private static void addLogProperties(Map config) {
        String worker = System.getProperty("logfile.name")?.replace(".log", "")
        if (worker) MDC.put("worker", worker)
        String topologyName = config.get("topologyName")
        if (topologyName) MDC.put("topologyName", topologyName)
    }

    private static dumpConfigProperties(Map config) {
        log.debug("System Properties:")
        System.properties.each {
            log.debug(it)
        }
        log.debug("Config Properties:")
        config.each {
            log.debug(it)
        }
    }

}
```

PR #603, opened 2013-06-21
0.9.0 windows

davidlao2k:
Add support for running Storm on Windows:
- use file path separator and class path separator to manipulate file and class paths
- add on-windows? to deal with windows specific logics
- use windows taskkill.exe to kill process
- add logics to deal with windows specific directory removal behavior

### DIFF
--- a/src/clj/backtype/storm/daemon/supervisor.clj
+++ b/src/clj/backtype/storm/daemon/supervisor.clj
@@ -99,11 +99,11 @@
      {}
      (dofor [[id hb] id->heartbeat]
             (let [state (cond
+                         (not hb)
+                           :not-started
                          (or (not (contains? approved-ids id))
                              (not (matches-an-assignment? hb assigned-executors)))
                            :disallowed
-                         (not hb)
-                           :not-started
                          (> (- now (:time-secs hb))
                             (conf SUPERVISOR-WORKER-TIMEOUT-SECS))
                            :timed-out
@@ -263,10 +263,18 @@
        (map :storm-id)
        set))
 
-(defn try-shutdown-workers [supervisor]
-  (let [worker-root (worker-root (:conf supervisor))
-         ids (read-dir-contents worker-root)]
-    (doseq [id ids]
+(defn shutdown-disallowed-workers [supervisor]
+  (let [conf (:conf supervisor)
+        ^LocalState local-state (:local-state supervisor)
+        assigned-executors (defaulted (.get local-state LS-LOCAL-ASSIGNMENTS) {})
+        now (current-time-secs)
+        allocated (read-allocated-workers supervisor assigned-executors now)
+        disallowed (keys (filter-val
+                                  (fn [[state _]] (= state :disallowed))
+                                  allocated))]
+    (log-debug "Allocated workers " allocated)
+    (log-debug "Disallowed workers " disallowed)
+    (doseq [id disallowed]
       (shutdown-worker supervisor id))
     ))
 
@@ -325,12 +333,12 @@
       ;; important that this happens after setting the local assignment so that
       ;; synchronize-supervisor doesn't try to launch workers for which the
       ;; resources don't exist
+      (if on-windows? (shutdown-disallowed-workers supervisor))
       (doseq [storm-id downloaded-storm-ids]
         (when-not (assigned-storm-ids storm-id)
           (log-message "Removing code for storm id "
                        storm-id)
           (try
-            (if on-windows? (try-shutdown-workers supervisor))
             (rmr (supervisor-stormdist-root conf storm-id))
             (catch Exception e (log-message (.getMessage e))))
           ))
--- a/src/clj/backtype/storm/daemon/supervisor.clj
+++ b/src/clj/backtype/storm/daemon/supervisor.clj
@@ -149,7 +149,10 @@
     (rmpath (worker-root conf id))
   (catch RuntimeException e
     (log-warn-error e "Failed to cleanup worker " id ". Will retry later")
-    )))
+    )
+  (catch java.io.FileNotFoundException e (log-message (.getMessage e)))
+  (catch java.io.IOException e (log-message (.getMessage e)))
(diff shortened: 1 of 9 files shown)


src/clj/backtype/storm/daemon/supervisor.clj
```diff
@@ -160,7 +163,9 @@
       (psim/kill-process thread-pid))
     (doseq [pid pids]
       (ensure-process-killed! pid)
-      (rmpath (worker-pid-path conf id pid))
+      (try
+        (rmpath (worker-pid-path conf id pid))
+        (catch Exception e)) ;; on windows, the supervisor may still holds the lock on the worker directory
```
nathanmarz, line 168: I don't quite understand... so in this case will it be cleaned up appropriately or not?

src/clj/backtype/storm/daemon/supervisor.clj
```diff
@@ -258,6 +263,21 @@
        (map :storm-id)
        set))
 
+(defn shutdown-disallowed-workers [supervisor]
```
nathanmarz, line 266: Can you explain this one?

src/clj/backtype/storm/testing.clj
```diff
@@ -32,7 +32,7 @@
   (FeederSpout. (Fields. fields)))
 
 (defn local-temp-path []
-  (str (System/getProperty "java.io.tmpdir") "/" (uuid)))
+  (str (System/getProperty "java.io.tmpdir") (if-not on-windows? "/") (uuid)))
```
nathanmarz, line 35: If it's on windows is the path separator already there?

Probably the best thing to do here is factor out a concat-paths function which explicitly handles / and \ appropriately, this seems hacky.

src/clj/backtype/storm/ui/core.clj
```diff
@@ -733,9 +733,11 @@
            ui-template))
   (GET "/topology/:id" [:as {cookies :cookies} id & m]
        (let [include-sys? (get-include-sys? cookies)]
-         (-> (topology-page id (:window m) include-sys?)
+         (try
+           (-> (topology-page id (:window m) include-sys?)
              (concat [(mk-system-toggle-button include-sys?)])
-             ui-template)))
+             ui-template)
+           (catch Exception e (resp/redirect "/")))))
```
nathanmarz, line 740: Why do you have to catch an exception here?

src/clj/backtype/storm/util.clj
```diff
@@ -429,7 +438,8 @@
 
 (defn touch [path]
   (log-debug "Touching file at " path)
-  (let [success? (.createNewFile (File. path))]
+  (let [success? (do (if on-windows? (.mkdirs (.getParentFile (File. path))))
```
nathanmarz, line 441: Is this a bug in Java that mkdirs doesn't work?

nathanmarz:
Sorry for taking so long on this. I made a lot of comments/questions inline, here's more general comments:
1. There's a lot of exception catching going on. Is it possible to implement these operations in a different way so that this isn't necessary? Having all that exception-catching is really dirty. At the very least, it should rethrow the exception unless it's in Windows.
2. Why is the bin/storm.cmd script necessary? Why can't we use bin/storm and modify it appropriately to work on Windows?

davidlao2k:
Hi Nathan,
Thanks for getting back to me. I'll take a look at your other questions and
comments and get back to you soon. With respect to 2) it is mainly to aid
our production deployment where c python isn't sanctioned. Also when
developing on windows it is very common to launch the service daemons in
separate console windows so one can the traces in progress. For 1) I'll
review the exception cases again since I haven't look at these for a while.

Thanks,
-David

src/clj/backtype/storm/ui/core.clj
davidlao2k, line 740: This attempts to give user a more graceful ui transition when killing a
running topology with the "kill" bottom. Since the "/topology/:id" url is
no longer valid, refreshing can lead to a screen full of stack dump.  It's
more of a nice to have.

src/clj/backtype/storm/testing.clj
davidlao2k, line 35: Agreed. This should be handled by a path concat function. The java.nio.file
Path class could it but that requires java 7.  Is java backward
compatibility a must?

src/clj/backtype/storm/daemon/supervisor.clj
davidlao2k, line 266: Yes, this is a nasty one. On Windows when a process holds handle to a file,
that file can not be removed until the process frees it. So unless the
worker processes associate with the topology are shutdown the rmr operation
will fail.

src/clj/backtype/storm/daemon/supervisor.clj
davidlao2k, line 168: This prevents the supervisor from going down. I need to look into this
again but we may have added a out of band clean up job to clean up. I have
a few other things I need to follow up on I'll get back to you on this.
 Thanks Nathan.

nathanmarz:
I don't think not wanting to use a dependency justifies this kind of code duplication – the storm.cmd script can be maintained as a separate project for those for whom this is a problem. Also, as for launching the service daemons... the bin/storm commands are all blocking, so each daemon can easily be launched in a separate console window.

src/clj/backtype/storm/testing.clj
ptgoetz, line 35: I think there are a lot of users (including myself) that are running with Java 6. I think we should have a very compelling reason to force users to upgrade to 7.

If this can be addressed in Java 6, I'd prefer that.

ptgoetz:
@davidlao 

Any thoughts on the pull request comments?

Should we punt this to 0.9.1?

ptgoetz:
Moved to 0.9.1

ptgoetz:
Closing at the request of @davidlao2k. This pull request has been superseded by #722.

PR #606, opened 2013-06-24
Documentary comments for storm internals

mrflip:
Added documentary comments to the critical pieces of executor and acker code.


nathanmarz:
Documentation tends to become out of sync with the code, and it can become a maintenance nightmare. I'd rather we did this as a wiki page with links into the code, ala what I started doing here: https://github.com/nathanmarz/storm/wiki/Implementation-docs

Can you convert this to a wiki page?

PR #607, opened 2013-06-25
Fixes to build_release.sh script

mrflip:
- maven is putting the `storm-*` jar files in target/release+provided, not target/provided. Fixed path.
- storm-netty had a dependency itself on storm-core, so the storm-core jar was in both _release and _release/lib. Removing the dupe jar from `lib/`
- storm-console-processor was letting clojure-1.2 meet its dependency list, making for two clojure jars in the tarball. Added a dependency to its `project.clj`.
- build_releases.sh should fail if the build_modules.sh script fails. Made build_modules.sh run with set -e, and made build_releases.sh fail if it does
- made pretty banners, made paths be properly quoted
- bin/ directory files copy with permissions intact. Also now remove build_release.sh and build_modules.sh, since they won't work
- rather than remove _release and *.zip, only removes the specific subdirectory and .zip file. If you run with `STORM_KEEP_RELEASE=true`, it will not delete the release dir at the end of script (only at beginning)

### DIFF
--- a/bin/build_release.sh
+++ b/bin/build_release.sh
@@ -1,62 +1,87 @@
 #!/bin/bash
 function quit {
-    exit 1
+  exit 1
 }
-trap quit 1 2 3 15  #Ctrl+C exits.
+trap quit 1 2 3 15  # Ctrl+C exits.
+
+function banner {
+  echo ; echo "=========================" ; echo "==" ; echo "== $1" ; echo "==" ; echo
+}
 
 RELEASE=`cat VERSION`
-LEIN=`which lein2 || which lein` 
+LEIN=`which lein2 || which lein`
 export LEIN_ROOT=1
 
-echo Making release $RELEASE
+# ==========================================================================
+banner "Making release $RELEASE"
 
-DIR=`pwd`/_release/storm-$RELEASE
+CODE_ROOT=`pwd`
+DIR="$CODE_ROOT/_release/storm-$RELEASE"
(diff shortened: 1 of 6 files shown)


mrflip:
Just saw this overlaps with #597 -- let me know if I should rebase. (This patch implements the same functionality)

bin/build_release.sh
```diff
@@ -1,55 +1,80 @@
...
 for module in $(cat MODULES)
 do
-	cd $module
-	mvn dependency:copy-dependencies || exit 1
-	cp -f target/dependency/*.jar $DIR/lib/
-	cp -f target/*.jar $DIR/
-	cd ..
+  cd $module
+  mvn dependency:copy-dependencies || exit 1
+  cp -f target/release/dependency/*.jar "$DIR/lib/"
```
nathanmarz: For me these dirs don't exist, I just have target/dependency.

nathanmarz:
I tried running this and it doesn't work on my computer. No jars are copied to the release distribution.

mrflip:
Gah sounds like I'm gonna have to read some maven documentation. I'll update the pull request once I know why we get different destinations.

mrflip:
Looks like technomancy/leiningen@a1d7b6f3eabe4eb made it append the profile name to the target directory -- if you installed lein before May, you got a pom with `<directory>target</directory>`; with recent lein, `<directory>target/release</directory>`. 

I made the project.clj file fix the directory as just plain `target`, and rebased on top of current master.

PR #618, opened 2013-07-01
install_zmq.sh cleans up after itself

zurawiki:
wget -c makes sure that the file is written as zeromq-2.1.7.tar.gz, not zeromq-2.1.7.tar.gz.1, zeromq-2.1.7.tar.gz.2, ...

On success of untarring (&&), the archive is removed


nathanmarz:
Sorry, I'm not accepting pull requests to this script, it's just meant to be helpful. Ultimately 0mq is so finicky there's no way we can make this script work on every system.

PR #621, opened 2013-07-03
Passing configs from storm.yaml to TopologyValidator

ankitoshniwal:
Passing contents of storm.yaml to TopologyValidator so that we can add more hooks in topology validator. For example with this change we can say that only allow those topologies to be pushed which are part of the isolation scheduler map in yaml file.

Thanks,
Ankit.

### DIFF
--- a/storm-core/src/clj/backtype/storm/daemon/nimbus.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/nimbus.clj
@@ -907,7 +907,8 @@
           (.validate ^backtype.storm.nimbus.ITopologyValidator (:validator nimbus)
                      storm-name
                      (from-json serializedConf)
-                     topology)
+                     topology
+                     conf)
           (swap! (:submitted-count nimbus) inc)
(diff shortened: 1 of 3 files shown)


nathanmarz:
Can you change this to add a prepare method to ITopologyValidator that is given the storm.yaml conf? That would keep the interface consistent with how the rest of Storm works.

nathanmarz:
@ankitoshniwal I'd like to get this merged in. If you aren't planning on making these changes please close this pull request.

ankitoshniwal:
Hey @nathanmarz sorry for the delay, just updated the ITopologyValidator file. Please let me know if you have any concerns.

Thanks,
Ankit.

nathanmarz:
That's not what I meant. I don't think this will even compile. ITopologyValidator#prepare should only take in the nimbus conf, and the nimbus conf should not be passed into validate. Additionally, when the validator is created the prepare method needs to be called on it.

ankitoshniwal:
Ok cool, I will do the changes.

For now I will close this pull request and open another one maybe in a day or so after I have validated all the changes.

Thanks Nathan!!

Ankit.

PR #422, opened 2012-12-12
Nimbus HA

Frostman:
The second part of Nimbus HA implementation.

---

This change is [Reviewable](https://reviewable.io/reviews/nathanmarz/storm/422)

PR #622, opened 2013-07-06
enable nimbus & UI servers to be asisgned with available port, instead of pre-configured 

anfeng:
This is one of a series pull requests, which aim to launch Storm cluster in a cluster environment.

Currently, Nimbus server and UI server launched with a configured port number. In a cloud environment, 
we don't really know which port is actually available. So, it will be convenient if Nimbus & UI server could simply 
find an available port and launch it.

In this pull request, we enable Nimbus server and UI server to find ports dynamically:
- If the nimbus port is configured to be <=0, Nimbus server will find an available port.
- If the UI port is configured to be <=0, UI server will find an available port.
  Nimbus/UI server announces the port number via Zookeeper. 

When Supervisor and UI server starts, they will pick up Nimbus host/port from Zookeeper (instead of simply reading configuration file).

### DIFF
--- a/storm-core/src/clj/backtype/storm/bootstrap.clj
+++ b/storm-core/src/clj/backtype/storm/bootstrap.clj
@@ -39,7 +39,7 @@
                      KillOptions SubmitOptions RebalanceOptions JavaObject JavaObjectArg
                      TopologyInitialStatus]))
      (import (quote [backtype.storm.daemon.common StormBase Assignment
-                     SupervisorInfo WorkerHeartbeat]))
+                     SupervisorInfo WorkerHeartbeat HostPort]))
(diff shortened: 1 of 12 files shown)


nathanmarz:
This overlaps with the work done in #422 to implement HA Nimbus (it uses ZK for Nimbus discovery as part of the implementation). It would be great to get some more work done on that one (it just needs to be tested more).

anfeng:
Do you want to merge #422 first? This new pull request is much simpler though.

nathanmarz:
The HA Nimbus is far more important... and this issue would just create more work for that one. For now, for the cloud-specific use case, can't whatever's launching Nimbus find a port and configure it appropriately? That's effectively how the worker ports are configured in storm-mesos.

PR #625, opened 2013-07-13
Transition away from Java serialization for storing state on disk or in Zookeeper

mrflip:
This is all @hausdorff's work -- it addresses #419 and #525, and supersedes #497

I've rebased his commits into the current master, and validated that the tests pass.

Hopefully this helps grease the path for the PR to go in...


ptgoetz:
@mrflip @hausdorff 
Have you smoke tested this in a clustered environment?

If so I'm +1.

storm-core/src/clj/backtype/storm/cluster.clj
```diff
@@ -230,7 +230,7 @@
       (assignment-info [this storm-id callback]
         (when callback
           (swap! assignment-info-callback assoc storm-id callback))
-        (maybe-deserialize (get-data cluster-state (assignment-path storm-id) (not-nil? callback)))
+        (maybe-deserialize-clj-bytes (get-data cluster-state (assignment-path storm-id) (not-nil? callback)))
```
anfeng, line 233: Why are we renaming these functions? If we stick the original func name, we will have less code to worry about. No?

storm-core/src/clj/backtype/storm/daemon/common.clj
```diff
@@ -40,12 +40,6 @@
 (defprotocol DaemonCommon
   (waiting? [this]))
 
-(def LS-WORKER-HEARTBEAT "worker-heartbeat")
```
anfeng, line 43: Why do we want to move these constants to Constants.java? We have many other CLJ constants, and Constants only has a small subset. Personally, I would rather we have a separated PR on that. This PR should just focus on getting serialization issues addressed.

storm-core/src/jvm/backtype/storm/utils/Utils.java
```diff
@@ -71,6 +74,30 @@ public static Object deserialize(byte[] serialized) {
         }
     }
 
+    public static byte[] serializeTopology(StormTopology topo) {
```
anfeng, line 77: You have some format issue here. Please avoid TAB.

anfeng, line 77: Please introduce a new Java class, say TopologySerializer.  You could then create serializer and deserializer object once, instead for every invocation.

hausdorff:
Hey, cool -- good to see this finally getting resolved.

I'll be happy to take a look at this in a couple hours, after I get off work.

ptgoetz:
@hausdorff @mrflip 
Any update on this?

PR #626, opened 2013-07-15
Allow ShellBolts to optionally specify the logging level.

msukmanowsky:
Default logging will still be info to support backwards compatibility.

### DIFF
--- a/storm-core/src/jvm/backtype/storm/task/ShellBolt.java
+++ b/storm-core/src/jvm/backtype/storm/task/ShellBolt.java
@@ -61,6 +61,10 @@ public class ShellBolt implements IBolt {
     private Thread _readerThread;
     private Thread _writerThread;
 
+    private enum LogLevels {
+        TRACE, DEBUG, INFO, WARN, ERROR, FATAL
+    }
+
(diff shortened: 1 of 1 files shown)


nathanmarz:
Can you send me a contributor agreement? Instructions here: https://github.com/nathanmarz/storm/wiki/Contributing-to-Storm

msukmanowsky:
Sent!

nathanmarz:
I'm not a fan of this implementation. Can you use numbers instead of strings for specifying the level? 0 = debug, 1 = info, etc. And it seems like it should test against info first because that's the common case.

msukmanowsky:
Hmm, considering this is implemented in the ShellBolt, having an enum kind of implementation (0 = debug, 1 = info) means that each language would need to maintain the same kind of enum.  I thought it'd be an easier interface for the languages to specify the log level using a string no?

nathanmarz:
Numbers are more performant. This is a protocol, it should be done efficiently.

msukmanowsky:
Alright, think we're all set here, mind giving it a look @nathanmarz?

PR #633, opened 2013-07-21
Disruptor queue metrics (patch for #613)

mrflip:
- Added new metric type, 'StateMetric', to report direct metrics of anything that meets the IStatefulObject interface
- Added methods for population, capacity, and read/write head positions to Disruptor Queue; gave it a getState() to make it IStatefulObject
- Register the transfer queue and the send and receive queues for spouts and bolts.

We cheat and register the transfer queue in executor.clj because within a Java implementation of the bolt (specifically, the System bolt), we don't have access to the executor data -- we need a TopologyContext, which is only prepared for tasks. If there's a way to get this, please advise.

### DIFF
--- a/storm-core/src/clj/backtype/storm/daemon/executor.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/executor.clj
@@ -6,7 +6,8 @@
   (:import [backtype.storm.spout ISpoutWaitStrategy])
   (:import [backtype.storm.hooks.info SpoutAckInfo SpoutFailInfo
             EmitInfo BoltFailInfo BoltAckInfo BoltExecuteInfo])
-  (:import [backtype.storm.metric.api IMetric IMetricsConsumer$TaskInfo IMetricsConsumer$DataPoint])
+  (:import [backtype.storm.metric.api IMetric IMetricsConsumer$TaskInfo IMetricsConsumer$DataPoint StateMetric])
+  (:import [backtype.storm Config])
   (:require [backtype.storm [tuple :as tuple]])
   (:require [backtype.storm.daemon [task :as task]])
   (:require [backtype.storm.daemon.builtin-metrics :as builtin-metrics]))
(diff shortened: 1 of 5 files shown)


mrflip:
Does this need to ignore certain executors -- for example, metrics consumers? Or is that OK?

If it does, what should we test for?

storm-core/src/clj/backtype/storm/daemon/executor.clj
```diff
@@ -656,6 +659,16 @@
                                                                                (MessageId/makeId anchors-to-ids)))))
                                     (or out-tasks [])))]]
           (builtin-metrics/register-all (:builtin-metrics task-data) storm-conf user-context)
+          (if (= component-id Constants/SYSTEM_COMPONENT_ID)
```
nathanmarz, line 663: Why the special case?

nathanmarz:
There's a lot of reformatting in this pull request, making it really hard to read what changed. Please do formatting as separate pull requests.

Some questions:
1. What's the implication of reading those variables on a different thread than they're being used?
2. What about monitoring the worker queue that sends tuples to remote workers?

storm-core/src/clj/backtype/storm/daemon/executor.clj
mrflip, line 663: We couldn't figure out a better opportunity to grab the transfer queue. If it's the system bolt, it instruments the send, recv and transfer queues; otherwise it just adds send and receive. If there's a better lifecycle point please advise.

mrflip:
Fixed the whitespace. Will file a separate issue for global whitespace / formatting fixes.

When you say "the worker queue that sends tuples to remote workers", is that the transfer queue? If not, where should I find it?

The transfer queue is instrumented on lines 664+: If it's the system bolt, it instruments the send, recv and transfer queues; otherwise it just adds send and receive. If there's a better lifecycle point to do this please advise -- most of the obvious places lack a TopologyContext.

mrflip:
A few more thoughts:

**threads**: Since the variables are being read from a different thread, it's absolutely possible for the read or write head to advance between taking the value of the one and the value of the other. I believe however this will be very rare, and not impact the actionable quality of the value. I don't like the idea of adding synchronization near this.

I did update the PR to a) get each head position once and use them to find the current capacity; b) make sure to get the read and then the write so that it is never an under-estimate of the capacity (or negative).

**tasks**: As far as we can tell the metrics topology context is only available per-task -- and so this code makes one metrics consumer per _task_, not per _executor_. I can't tell if that's a good idea (since the other instruments are also per-task) or a bad idea (because the thing it's measuring is per-executor).

PR #640, opened 2013-08-01
Launch worker per environment settings

anfeng:
To launch storm workers in a cloud environment, we need to handle some minor issues:
- worker's logs have to be located in certain directory (instead of current working directory)
- logback configuration file is located in a distributed file system directory (instead of  current working directory)
- JAVA_HOME needs to respected if it's specified in an environment

This pull request makes the following challenges to supervisor's logic to launch workers:
- If JAVA_HOME is specified in an environment, we will use <JAVA_HOME>/bin/java
- If "worker.logdir" is specified in supervisor's command (-Dworker.logdir=xxxx), worker logs will be stored in that directory
- worker's logback configuration file will be identical to supervisor's logback configuration.


storm-core/src/clj/backtype/storm/daemon/supervisor.clj
```diff
@@ -412,12 +412,17 @@
           childopts (.replaceAll (str (conf WORKER-CHILDOPTS) " " (storm-conf TOPOLOGY-WORKER-CHILDOPTS))
                                  "%ID%"
                                  (str port))
-          logfilename (str "worker-" port ".log")
-          command (str "java -server " childopts
+          java_home (System/getenv "JAVA_HOME")
+          java_cmd (if java_home (str java_home "/bin/java") "java")
+          worker_logdir (System/getProperty "worker.logdir")
+          logfilename (str (if worker_logdir (str worker_logdir "/") "") "worker-" port ".log")
```
d2r, line 418: Would it be good to use the `File/separator` instead of "/" while we are changing this?

PR #651, opened 2013-08-12
Improve util/join-maps to remove nils 

fxjwind:
(join-maps {"c1" 2} {"c2" 1} {"c3" 3})
current, {c1 (2 nil nil), c2 (nil 1 nil), c3 (nil nil 3)}
my fix, {c1 (2), c2 (1), c3 (3)}

I also tryed other way, but the sequence of map items will be change.

``` Clojure
(defn join-maps [& maps]
"({:a 1 :b 2 :c 3} {:a 2 :c 4})->{:b (2), :c (3 4), :a (1 2)}" 
  ( map-val seq
    (reduce (fn [m map] 
            (reduce (fn [m [k,v]] (assoc m k (conj (get m k []) v))) m map))
    {} maps))
)
```

So I just remove the nils on oringal logic finally

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.project
 .lein-repl-history
 /classes
 /lib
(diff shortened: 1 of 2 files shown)


nathanmarz:
This is not the intention of that function.

PR #652, opened 2013-08-15
Errors during deserialization of spouts/bolts should go to the Storm UI

mvogiatzis:
This should fix issue nathanmarz/storm#358.

This is the first time I commit code to storm repository - I have sent you the contributor agreement filled and signed. Let me know if anything else is needed.

Cheers,
Michael
@mvogiatzis

### DIFF
--- a/storm-core/src/clj/backtype/storm/daemon/task.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/task.clj
@@ -162,8 +162,8 @@
     :user-context (user-topology-context (:worker executor-data) executor-data task-id)
     :builtin-metrics (builtin-metrics/make-data (:type executor-data))
     :tasks-fn (mk-tasks-fn <>)
-    :object (get-task-object (.getRawTopology ^TopologyContext (:system-context <>)) (:component-id executor-data))))
-
+    :object (try (get-task-object (.getRawTopology ^TopologyContext (:system-context <>)) (:component-id executor-data)))
+    (catch RuntimeException re (:report-error-and-die executor-data))))
 
 (defn mk-task [executor-data task-id]
   (let [task-data (mk-task-data executor-data task-id)
--- a/storm-core/src/clj/backtype/storm/daemon/task.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/task.clj
@@ -163,7 +163,7 @@
     :builtin-metrics (builtin-metrics/make-data (:type executor-data))
     :tasks-fn (mk-tasks-fn <>)
     :object (try (get-task-object (.getRawTopology ^TopologyContext (:system-context <>)) (:component-id executor-data)))
-    (catch RuntimeException re (:report-error-and-die executor-data))))
+    (catch Exception e (:report-error-and-die executor-data))))
 
 (defn mk-task [executor-data task-id]
   (let [task-data (mk-task-data executor-data task-id)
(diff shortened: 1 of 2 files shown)


nathanmarz:
I think this should catch all Throwables, not just RuntimeExceptions. It's going to die anyway, so it might as well make a best effort to report anything that goes wrong.

mvogiatzis:
I agree, it should catch everything. Just committed.

storm-core/src/clj/backtype/storm/daemon/task.clj
```diff
@@ -162,8 +162,8 @@
     :user-context (user-topology-context (:worker executor-data) executor-data task-id)
     :builtin-metrics (builtin-metrics/make-data (:type executor-data))
     :tasks-fn (mk-tasks-fn <>)
-    :object (get-task-object (.getRawTopology ^TopologyContext (:system-context <>)) (:component-id executor-data))))
-
+    :object (try (get-task-object (.getRawTopology ^TopologyContext (:system-context <>)) (:component-id executor-data)))
+    (catch Exception e (:report-error-and-die executor-data))))
```
ptgoetz, line 166: `RuntimeException` (unchecked exception) and `Exception` (checked exception) both extend `Throwable`. So by just catching `Exception` you ignore all `RuntimeException`s.

Change this to catch `Throwable` and I'm +1

revans2, line 166: Throwable has two subclasses Error and Exception.  RuntimeException is a subclass of Exception.  Both Error and RuntimeException are unchecked, however Errors indicate that something very seriously wrong has happened and catching it could possibly leave the JVM in an unknown state, Think OutOfMemoryError.  In practice it is not often that bad, but I would rather avoid it unless I have to.

All of the Event manager threads and server threads already catch Throwable and kill the process if it makes it up to that point.

Plus the code is not even reporting the error properly

``` clojure
(:report-error-and-die executor-data) 
```

is simply pulling the error reporting function out of the executor data.  It is not executing it in any way.  If this happens :object will be set to the error reporting function, and then really odd behavior will start to happen.

mvogiatzis, line 166: I'm not a Clojure expert but I found this [article](http://www.javatuning.com/why-catch-throwable-is-evil-real-life-story/) which is relevant to what @revans2 just mentioned.

ptgoetz, line 166: Thanks for correcting me, you're right. 

Nonetheless, it would be nice to have the error reported in the ui, even if it is something like an OOME. 
Would you have a suggestion or patch to make this right?

We'd like to include this functionality in 0.9.0

Thanks!

Taylor

ptgoetz, line 166: @mvogiatzis Do @revans2 's comments make sense?

You are on the right track, but the function isn't actually getting invoked.

If the error reporting function is rethrowing, then catching throwable should be okay (since it would then bubble up the stack).

I'll try to take a closer look when I have better access to the source code (not on a phone).

xumingming, line 166: Agree with @ptgoetz that the function is not invoked: `(:report-error-and-die executor-data)` just returns the function, but not invoked. 

For the decision between `Throwable` and `Exception`, I personally prefer `Exception`, it's safer. If we use `Throwable`, we need to rely on `report-error-and-die` to do the right thing.

ptgoetz:
@mvogiatzis 

Looks like there is consensus on using `Exception`. 

All that seems to remain is to fix the code so that the function returned by `(:report-error-and-die executor-data)` gets executed.

Once that's done I'm +1.

Let me know if you need any help with that.

PR #658, opened 2013-08-20
Filter empty args before passing to os.execvp.

stevensurgnier:
This PR is an attempt to fix https://github.com/nathanmarz/storm/issues/657.

A `java.lang.NoClassDefFoundError` exception is thrown if `JAR_JVM_OPTS` is empty. This PR filters empty arguments before passing `all_args` to `os.execvp`.  The functionality intended by https://github.com/nathanmarz/storm/commit/0fca3034ff90c8d5de2e327d58085580fcc74256 should still remain.

### DIFF
--- a/bin/storm
+++ b/bin/storm
@@ -112,13 +112,13 @@ def parse_args(string):
 
 def exec_storm_class(klass, jvmtype="-server", jvmopts=[], extrajars=[], args=[], fork=False):
     global CONFFILE
-    all_args = [
+    all_args = filter(None, [
         "java", jvmtype, get_config_opts(),
         "-Dstorm.home=" + STORM_DIR, 
         "-Djava.library.path=" + confvalue("java.library.path", extrajars),
         "-Dstorm.conf.file=" + CONFFILE,
         "-cp", get_classpath(extrajars),
-    ] + jvmopts + [klass] + list(args)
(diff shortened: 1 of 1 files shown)


bin/storm
```diff
@@ -139,7 +139,7 @@ def jar(jarfile, klass, *args):
         jvmtype="-client",
         extrajars=[jarfile, USER_CONF_DIR, STORM_DIR + "/bin"],
         args=args,
-        jvmopts=[JAR_JVM_OPTS + " -Dstorm.jar=" + jarfile])
+        jvmopts=[JAR_JVM_OPTS, "-Dstorm.jar=" + jarfile])
```
mrflip, line 142: I think it's better to do the filter here (`filter(None, [JAR_JVM_OPTS, "-Dstorm.jar=" + jarfile])`), rather than on the all_args list. 

Nothing else in `all_args` should be null; applying that filter to `all_args` would make it quietly proceed rather than break.

stevensurgnier, line 142: The only concern I have with your proposal is that there are no tests for `/bin/storm` to guarantee the semantics that every argument in `all_args` **must** not be null. Hence, how the bug was introduced in the first place. By filtering `all_args` we can guarantee the semantics in the absence of tests.

xumingming:
Another merged PR have already done the same thing: #683

PR #662, opened 2013-08-23
May be a little bug in supervisor 

wugang33:
When i test supervisor .First I kill the supervisor and worker. And when the nimbus remove the assignment from this supervisor .I start  the supervisor.I got a Exception 
2013-08-23 15:45:16 event [ERROR] Error when processing event
java.io.FileNotFoundException: File 'storm-local/supervisor/stormdist/test-1-1377241347/stormconf.ser' does not exist

I think there is a bug,when the synchronize-supervisor and sync-processes execute at one time ,the sync-processes get the old assignment from the local-state when synchronize-supervisor not already put new assignment in the local-state.


nathanmarz:
This is fixed in 0.9.0 and 0.8.3 wips. If you think not, then open an issue first.

PR #663, opened 2013-08-26
make FixedTupleSpout implement IRichSpout so it can be used for tests

jaked:
The Summingbird guys want to use FixedTupleSpout explicitly in tests (rather implicitly via the :mock-sources argument to complete-topology).


sritchie:
nice, thanks @jaked.

sritchie:
Hey @nathanmarz, any plans on a new wip release? I see the latest Clojars came out in july. It'd be great if you could cut a new release so we could pull this code into Tormenta.

PR #667, opened 2013-08-28
Ignoring a FileNotFound excpetion that can bring down a supervisor.

revans2:
When running tests we found a race. kill_local_storm_cluster in testing.clj tries to shut down the supervisor and kill all running workers using try_worker_cleanup.  At the same time a worker may have exited and the sync_process would also call try_worker_cleanup.  The last one called gets a FileNotFoundException from the call to rmr.  rmr can also throw an IOException if there are other issues with deleting the directory.  I thought it was best just to ignore this error like we do RuntimeExceptions thrown by rmpath.


nathanmarz:
Wouldn't it be better to catch and suppress FileNotFoundExceptions in rmpath itself? The ultimate goal of the function is to ensure that path is deleted, so it's fine to suppress that error.

revans2:
Will do.

PR #669, opened 2013-08-28
ui and log viewer send proper html pages

d2r:
Currently the response body starts with `<head>`.  Using `html4` will bookend the document with the proper tags and declarations.


mrflip:
+1

PR #671, opened 2013-09-03
Wrapping all curator exceptions in RuntimeExceptions.

revans2:
Recently we ran into a situation where an InterruptedException was being thrown from IOutputCollector.reportError, even though InterruptedException is not a RuntimeException.  Many of the Curator APIs actually declare that any Exception can be thrown.  To avoid non-runtime exceptions being thrown by a java API that does not declare it, this patch wraps any Exception thrown by the curator API in a RuntimeException.


nathanmarz:
Can we change this to only wrap the exception in a RuntimeException if it's not already a RuntimeException?

revans2:
Done.

PR #672, opened 2013-09-03
Update nimbus.clj

ankitoshniwal:
screen shot 2013-09-03 at 11 36 45 am

Adding check to disallow topology submission if the topology name is blank. In one of the scenarios, we had pushed a topology with an empty name, and it showed up on the ui with only the uid. Also since it did not have any name we could not go into "Topology Details" to kill it, and hence had to clean up the state on the disk so that Nimbus kills the topology.


nathanmarz:
Looks fine but can you fix the formatting? That if statement should be aligned with the other if statement.

ankitoshniwal:
Thanks Nathan!! Just updated the formatting.

Ankit.

nathanmarz:
One other thing – you don't need to wrap the error message in (str ...), as it's already a string.

PR #681, opened 2013-09-17
fixes #551, when rebalance ,remove out of date assignment

vinceyang:
to fixes #551 ,when rebalance the rebalance topolgy's assignment has out of date ,must clear these assignment from zookeeper,.

### DIFF
--- a/storm-core/src/clj/backtype/storm/cluster.clj
+++ b/storm-core/src/clj/backtype/storm/cluster.clj
@@ -117,6 +117,7 @@
   (update-storm! [this storm-id new-elems])
   (remove-storm-base! [this storm-id])
   (set-assignment! [this storm-id info])
+  (remove-assignment! [this storm-id])
   (remove-storm! [this storm-id])
   (report-error [this storm-id task-id error])
   (errors [this storm-id task-id])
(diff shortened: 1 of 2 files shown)


revans2:
This does seem like it will fix the re-balancing issue, but it does not address other schedulers that might take all resources from a previously running topology.

herberteuler:
My reference was a mistake.  Sorry for that.

vinceyang:
@revans2 ,  scheduler address all resources from a previously running topology,.seems we have to  mark all topology's a flag to describe does topology in effect. supervisor Ignore  the renbalancing topology's assigment. do you   have any idea?

revans2:
@vinceyang I am not really sure what you mean by that.  The scheduler looks at the cluster as a whole and all topologies that have been requested to run on it.  It has the power to remove or change assignments for any topology as it sees fit. I found that when a scheduler removes all of the assignments for a previously running topology, the code that decides if it should update ZK makes a mistake and does not update ZK.

I simply pointed out that the current patch seems to fix the case for a rebalance that can cause a scheduler to remove the resources for a previously running topology, but it does not address the general case where a scheduler might do the same thing for a different reason.  Assume that there is a priority queue and a higher priority topology comes in so the lowest priority topology is stopped.

PR #682, opened 2013-09-19
Two patches to scheduler

herberteuler:
There are two patches here:
1. There seems to be an erroneous structure in `default-schedule`: invoking to `schedule-topologies-evenly` should only be done **after** freeing all `bad-slots` for topologies that needs re-scheduling.
2. `SchedulerAssignmentImpl` currently uses `HashMap`, which **does not** preserve entries order.  This can cause unnecessary and unexpected re-scheduling when nimbus converts `executor->node+port` between its internal Clojure maps and `SchedulerAssignmentImpl`'s Java maps back and forth.

### DIFF
--- a/storm-core/src/clj/backtype/storm/scheduler/DefaultScheduler.clj
+++ b/storm-core/src/clj/backtype/storm/scheduler/DefaultScheduler.clj
@@ -55,8 +55,8 @@
                                     (not= alive-executors all-executors))
                                 (bad-slots alive-assigned (count all-executors) total-slots-to-use)
                                 [])]]
-      (.freeSlots cluster bad-slots)
-      (EvenScheduler/schedule-topologies-evenly (Topologies. {topology-id topology}) cluster))))
+      (.freeSlots cluster bad-slots))
+    (EvenScheduler/schedule-topologies-evenly (Topologies. {topology-id topology}) cluster)))
 
 (defn -schedule [this ^Topologies topologies ^Cluster cluster]
   (default-schedule topologies cluster))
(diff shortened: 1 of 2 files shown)


storm-core/src/clj/backtype/storm/scheduler/DefaultScheduler.clj
```diff
@@ -55,8 +55,8 @@
                                     (not= alive-executors all-executors))
                                 (bad-slots alive-assigned (count all-executors) total-slots-to-use)
                                 [])]]
-      (.freeSlots cluster bad-slots)
-      (EvenScheduler/schedule-topologies-evenly (Topologies. {topology-id topology}) cluster))))
+      (.freeSlots cluster bad-slots))
+    (EvenScheduler/schedule-topologies-evenly (Topologies. {topology-id topology}) cluster)))
```
xumingming, line 59: `topology-id` is declared in `doseq`, while you are referencing topology-id out of `doseq`

herberteuler, line 59: Okay, I see.

xumingming:
@herberteuler any update on this PR?

herberteuler:
The LinkedHashMap change has been in the branch for our testing cluster for a while, and it has fixed the unexpected rescheduling issue there.  We are planning to incorporate it into our next installation on the production cluster.

xumingming:
@herberteuler can you describe a case which have issues with current implementation, while is fixed by your patch?

herberteuler:
I'll make a new PR later, with more detailed descriptions.  Sorry for the inconvenience.

PR #683, opened 2013-09-24
Fix for STORM_JAR_JVM_OPTS which is currently broken when STORM_JAR_JVM_OPTS is not specified

roadkill001:


ptgoetz:
+1

anfeng:
+1

bin/storm
```diff
@@ -29,7 +29,7 @@ if (not os.path.isfile(USER_CONF_DIR + "/storm.yaml")):
     USER_CONF_DIR = CLUSTER_CONF_DIR
 CONFIG_OPTS = []
 CONFFILE = ""
-JAR_JVM_OPTS = os.getenv('STORM_JAR_JVM_OPTS', '')
+JAR_JVM_OPTS = os.getenv('STORM_JAR_JVM_OPTS' + " ", '')
```
xumingming: Why we need to append the extra space to `STORM_JAR_JVM_OPTS` when `getenv`?

ptgoetz: See the next diff...

It's really just moving it and making it conditional. If STORM_JAR_JVM_OPTS is undefined no space will be added. AFAICT.

xumingming: but the space should be added to the return value of `os.getenv` rather than the env name: `STORM_JAR_JVM_OPTS` right?

d2r: @xumingming agreed

Would it be better or worse to construct `jvmopts` by joining `JAR_JVM_OPTS` and `"-Dstorm.jar=" + jarfile` on a space?

```
jvmopts=[' '.join(filter(None, [JAR_JVM_OPTS, "-Dstorm.jar=" + jarfile]))]
```

ptgoetz: @d2r @xumingming yep, you're both absolutely right. I'm not reading code well today. ;)

@roadkill001 would you mind trying out @d2r's suggestion and amending the pull request?

nathanmarz:
@roadkill001 Also, we need a contributor agreement from you in order to merge this in. See here for instructions: https://github.com/nathanmarz/storm/wiki/Contributing-to-Storm

bin/storm
ptgoetz: I can vouch for @xumingming's approach. Changing line 142 to this:

``` python
jvmopts=[' '.join(filter(None, [JAR_JVM_OPTS, "-Dstorm.jar=" + jarfile]))])
```

Fixed the issue for me.

I'd prefer to have @roadkill001 amend the pull request and merge that, since @roadkill001 first discovered it and got the ball rolling for a fix.

But if we feel it's urgent I can add a new pull request so we wouldn't have to wait for a CLA.

roadkill001: I'll amend request... gimme few minutes to test

roadkill001: Checked in and ready

nathanmarz:
+1, and I've received a contributor agreement

PR #692, opened 2013-10-01
Adds macro while-timout to avoid test hangs

d2r:
Some tests have been getting stuck forever while waiting for a particular state in a test cluster.

This change will have tests error after 5 seconds in these situations instead of hanging indefinitely.


nathanmarz:
+1

xumingming:
+1

ptgoetz:
+1

I've experienced this behavior as well... So I'm all for the timeout to prevent hangs.

But in the long run I'd prefer to get to the root cause and have deterministic tests.

PR #693, opened 2013-10-01
Add in the ability to configure the number of worker threads for Netty...

revans2:
and set up some reasonable defaults for Netty.

In trying to find a good set of defaults for Netty I discovered that using the default number of worker threads caused a lot of contention due to excess context switching.  This patch adds in the ability to configure the number of worker threads Netty uses and sets some reasonable defaults that for my tests let Netty send between 40% and 110% more messages per second than zeromq.

This does not enable Netty by default, it just adds default configurations for when Netty is configured.

### DIFF
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -82,6 +82,13 @@ zmq.threads: 1
 zmq.linger.millis: 5000
 zmq.hwm: 0
 
+storm.messaging.netty.server_worker_threads: 1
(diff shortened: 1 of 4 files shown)


nathanmarz:
+1. I think this can go into 0.9.0 too.

xumingming:
+1

ptgoetz:
+1 for 0.9.0

PR #694, opened 2013-10-02
Add pluggable servlet context filter to ui

d2r:
This is useful to enable authentication of users to the UI with the
appropriate configured filter.


d2r:
We are using this with a filter that redirects to a single-sign-on to enforce user authentication.

There are several more pieces we have in place to make this useful, but I wanted to avoid a huge pull request and also put this up in light of #452, for which it may serve as a pointer to how something like https or http auth could be done.

d2r:
Closed on accident, reopening

d2r:
Added to [apache/storm](https://github.com/apache/storm/tree/ced622433eaff81d7039d384c3acf5de595b0a70)

PR #697, opened 2013-10-03
Make shell component serialisation plugable

jsgilmore:
This pull request makes multilang components plugable (issue #373) and relates to issue #654, in that it is now possible to implement protocol buffer serialisation.

Changes in this pull request:
Moved multilang classes to multilang directory.
Added BoltMsg, ShellMSg and SpoutMsg objects. These objects are used when communicating with a serializer, instead of raw JSON.
All JSON serialisation has been moved to a new JsonSerializer.
A serialiser is selected when creating a new ShellBolt or ShellSpout, by passing it in to the constructor. A default option of no serialiser automatically selects the built-in JsonSerializer.


xumingming:
change the package of `ShellBolt` makes this PR a breaking change: user code need to be modified to use the new package, is it really necessary?

jsgilmore:
It's not necessary, no. It was an effort to combine code that (to my mind at least) belongs together. If I had not created the multilang package, something like the utils package would be full of multilang code that both the ShellBolt and ShellSpout use. I suppose I could still create the multilang package and place all new code in there and leave the old code where it was. It means things are a bit all over the place, but at least that won't break anything. I do see these fragmented packages becoming unwieldy in the future.

xumingming:
Yeah, understand your concern. I personally like the other option your suggested: all new code goes into the new package.

jsgilmore:
I've now moved the existing code back to where it was and updated the dependancies. With the new code in the new package.

project.clj
```diff
@@ -10,7 +10,7 @@
                  :archive "https://groups.google.com/group/storm-user"
                  :post "storm-user@googlegroups.com"}
   :dependencies [~@DEPENDENCIES]
-  :plugins [[~'lein-sub "0.2.1"]]  
+  :plugins [[~'lein-sub "0.2.1"],[no-man-is-an-island/lein-eclipse "2.0.0"]]
```
xumingming: better keep the plugin config `lein-eclipse` in your own `profiles.clj` (`lein-sub` is configured here because it is kinda of a must due to current storm's multi-module structure)

storm-core/src/clj/backtype/storm/bootstrap.clj
```diff
@@ -17,7 +17,7 @@
      (import (quote [backtype.storm.task IBolt IOutputCollector
                      OutputCollector TopologyContext ShellBolt
                      GeneralTopologyContext WorkerTopologyContext]))
-     (import (quote [backtype.storm.coordination CoordinatedBolt CoordinatedBolt$SourceArgs
```
xumingming, line 35: remove white space.

xumingming:
better add some comments for the new codes, e.g. `Emission`, `Immission`, it makes it easier for the future maintainer to understand the code.

jsgilmore:
Will do. Referring to what @francis-liberty mentioned, that there might be other message types, already internal to Storm, which implement the fields of Immission and Emission? Could @xumingming  comment on that? Perhaps the tuple message type? I would prefer to use internal message types, if they already exist.

storm-core/src/clj/backtype/storm/bootstrap.clj
jsgilmore, line 35: Sorry, I couldn't find white space in the new code.

project.clj
jsgilmore: Fixed

storm-core/src/jvm/backtype/storm/multilang/JsonSerializer.java
```diff
@@ -0,0 +1,147 @@
...
+	private DataOutputStream processIn;
+    private BufferedReader processOut;
+
+	public void initialize (OutputStream processIn, InputStream processOut) {
+		this.processIn = new DataOutputStream(processIn);;
+		this.processOut = new BufferedReader(new InputStreamReader(processOut));
+	}
+
+	public Number connect (Map conf, TopologyContext context) throws IOException, NoOutputException {
+		JSONObject setupInfo = new JSONObject();
```
xumingming: seems the code is not well formatted, used `tab` in code?

storm-core/src/jvm/backtype/storm/spout/ShellSpout.java
```diff
@@ -18,88 +21,82 @@
     private SpoutOutputCollector _collector;
     private String[] _command;
     private ShellProcess _process;
+    private SpoutMsg spoutMsg;
```
xumingming, line 38: we should have three `SpoutMsg` object to represents the three message types: `fail`, `ack`, `next` commands(rather than just use a single one) -- these messages can come concurrently.

xumingming:
A new config item should be added to `Config.java` and `defaults.yaml`, so developer can specify a custom serializer.

storm-core/src/jvm/backtype/storm/utils/ShellProcess.java
```diff
@@ -1,85 +1,79 @@
...
 import java.util.List;
+import java.util.Map;
 
 import org.apache.commons.io.IOUtils;
-import org.json.simple.JSONObject;
-import org.json.simple.JSONValue;
 import org.apache.log4j.Logger;
 
-public class ShellProcess {
+public class ShellProcess implements Serializable {
```
xumingming, line 37: Why we need to implement `Serializable`?

storm-core/src/jvm/backtype/storm/spout/ShellSpout.java
```diff
@@ -18,88 +21,82 @@
...
-    
+
     public ShellSpout(String... command) {
         _command = command;
+        _process = new ShellProcess(new JsonSerializer(), _command);
+    }
+
+    public ShellSpout(ISerializer serializer, String... command) {
+        _command = command;
+        _process = new ShellProcess(serializer, _command);
```
xumingming: `new ShellProcess` might need to move to `open` method, so we can first get user configured serializer from `stormConf`, and pass the serializer to the `ShellProcess` constructor

storm-core/src/jvm/backtype/storm/spout/ShellSpout.java
```diff
@@ -18,88 +21,82 @@
...
+        _process = new ShellProcess(new JsonSerializer(), _command);
     }
-    
+
     public ShellSpout(String... command) {
         _command = command;
+        _process = new ShellProcess(new JsonSerializer(), _command);
+    }
+
+    public ShellSpout(ISerializer serializer, String... command) {
```
xumingming: This constructor seems not needed, since we will get the serializer class from `stormConf`(rather than let developer specify it in code).

storm-core/src/jvm/backtype/storm/spout/ShellSpout.java
```diff
@@ -18,88 +21,82 @@
...
 
     public ShellSpout(ShellComponent component) {
         this(component.get_execution_command(), component.get_script());
+        _process = new ShellProcess(new JsonSerializer(), _command);
     }
-    
+
     public ShellSpout(String... command) {
         _command = command;
+        _process = new ShellProcess(new JsonSerializer(), _command);
```
xumingming: hardcoded `JsonSerializer` should not be here, we might have a `default serializer` specified in `defaults.yaml`

storm-core/src/jvm/backtype/storm/task/ShellBolt.java
```diff
@@ -57,54 +56,53 @@
...
         this(component.get_execution_command(), component.get_script());
+        _process = new ShellProcess(new JsonSerializer(), _command);
     }
 
     public ShellBolt(String... command) {
         _command = command;
+        _process = new ShellProcess(new JsonSerializer(), _command);
+    }
+
+    public ShellBolt(ISerializer serializer, String... command) {
```
xumingming: just like `ShellSpout`: 
- this constructor is not needed
- `new ShellProcess` should be moved to `prepare` method
- hardcoded `JsonSerializer` should go to `defaults.yaml`

storm-core/src/jvm/backtype/storm/multilang/JsonSerializer.java
jsgilmore: I did. Should I use spaces? Are there any guidelines available for how you guys format your code?

storm-core/src/jvm/backtype/storm/spout/ShellSpout.java
jsgilmore, line 38: I wasn't aware of that. What would then happen if two ack messages come concurrently? Otherwise, how does Storm know when a spout has finished processing an ack?

storm-core/src/jvm/backtype/storm/multilang/JsonSerializer.java
xumingming: generally use 4 spaces to indent rather than a tab.

storm-core/src/jvm/backtype/storm/spout/ShellSpout.java
xumingming, line 38: I mean a `fail` message and a `next` message can come concurrently, so we can not use a single object to represent both of them.

storm-core/src/jvm/backtype/storm/spout/ShellSpout.java
```diff
@@ -18,88 +21,82 @@
...
-                    List<Object> tuple = (List) action.get("tuple");
-                    Object messageId = (Object) action.get("id");
-                    if (task == null) {
+                    String stream = emission.getStream();
+                    Long task = emission.getTask();
+                    List<Object> tuple = emission.getTuple();
+                    Object messageId = emission.getId();
+                    if (task == 0) {
                         List<Integer> outtasks = _collector.emit(stream, tuple, messageId);
-                        Object need_task_ids = action.get("need_task_ids");
```
xumingming, line 116: this `need_task_ids` logic is removed?

storm-core/src/jvm/backtype/storm/task/ShellBolt.java
```diff
@@ -57,54 +56,53 @@
...
         // reader
         _readerThread = new Thread(new Runnable() {
             public void run() {
                 while (_running) {
                     try {
-                        JSONObject action = _process.readMessage();
-                        if (action == null) {
-                            // ignore sync
-                        }
+                        Emission emission = _process.readEmission();
```
xumingming: is it possible that the `emission` is null, just like the old `action`?

storm-core/src/jvm/backtype/storm/task/ShellBolt.java
```diff
@@ -113,19 +111,20 @@ public void run() {
...
         });
-        
+
         _readerThread.start();
 
         _writerThread = new Thread(new Runnable() {
             public void run() {
                 while (_running) {
                     try {
+                    	// FIXME: This can either be TaskIds or Immissions
```
xumingming: this `FIXME` not fixed?

storm-core/src/jvm/backtype/storm/utils/ShellProcess.java
jsgilmore, line 37: When nimbus has to serialize the ShellProcess class to send it to supervisors, it wants it declared as serializable. I know it wasn't declared as serializable before and the only reason I can think of why it's needed now is because it contains an interface as member.

All I know is that topology builder wouldn't allow me to submit a topology. This was under Java 7, if that makes a difference.

xumingming:
a minor comment: are the terms `Emission`, `Immission` used frequently in your code? seems not very easy to understand at first glance(my feeling may be wrong, since English is not my native language).

storm-core/src/jvm/backtype/storm/spout/ShellSpout.java
jsgilmore: This sounds like a much better solution. I've not worked with stormConf like that before. Will look into how to do that.

storm-core/src/jvm/backtype/storm/spout/ShellSpout.java
jsgilmore, line 116: I removed this logic, since nowhere in the multilang protocol specification is this specified as an option (https://github.com/nathanmarz/storm/wiki/Multilang-protocol). I can put it back, which will just require an additional field to the Emission object. We then just have to update the multilang specification as well, which I won't mind doing.

storm-core/src/jvm/backtype/storm/task/ShellBolt.java
jsgilmore: So I looked at this code, and I could find no reason for action to be null as a consequence of a sync message, unless sync were sent only as "end" messages in the past. _process.readMessage (as it stands currently in the Storm code), reads a string and then parses it. If you look at readString, if only an "end" was received readString will return an empty string, which will translate to a null message. But for a sync message of the form: {"command":"sync"}, I can see no reason why action would be null.

The reason why I think this has not affected the use of shell components, is that if a valid action message were received (in the JSON format specified in the multilang protocol), it would also just be ignored, since the "sync" command is never handled in the if else over the "command" variable.

storm-core/src/jvm/backtype/storm/task/ShellBolt.java
jsgilmore: Ah, sorry. The if ... instaceof ... fixed that.

jsgilmore:
An emission is something that is received from somewhere, while an immission is something that is sent to somewhere. Admittedly, immission is not a word that is regularly used, although it is a perfectly good English word. It is also technically the opposite of emission, which is why I chose it. As I mentioned earlier, if there are other messages, already part of Storm, that contain the fields of emission and immission, I would rather use them.

If you can think of better names for them, I'm also open to that :-)

storm-core/src/jvm/backtype/storm/spout/ShellSpout.java
jsgilmore, line 38: Looking at the ShellSpout implementation, I can't see how this can work. If a spout receives a next message, it calls querySubprocess(). This function is a loop that sends a single message to the non-JVM program attached to it and then waits to receive messages until the non-JVM program sends a sync.

If next and ack can be received concurrently from Storm, it means that there can be two execution threads in the querySubprocess function at the same time. This would mean that the non-JVM program can receive an ack message, even while it is sending messages as a response to a next message. It is explicitly stated in the Storm multilang protocol that this can't happen.

xumingming, line 38: If I understand it correctly, this issue is not multilang protocol related. `ShellSpout` is a normal Storm spout, so the method `fail`, `nextTuple`, `ack` may be invoked by storm-core concurrently, so it seems that we need to synchronize the call to method `querySubprocess`? how do you think @nathanmarz ?

francis-shuoch:
@jsgilmore the class I recommended to use is Tuple. Here is the link to the [interface](https://github.com/nathanmarz/storm/blob/master/storm-core/src/jvm/backtype/storm/tuple/Tuple.java) and a storm provided [base class](https://github.com/nathanmarz/storm/blob/master/storm-core/src/jvm/backtype/storm/tuple/TupleImpl.java). I think it already has the fields you created in Emission and Immission. You might take a look. ;)

xumingming:
@francis-liberty `Tuple` in storm is designed to represent the messages transfered between Spouts and Bolts. I think it's not appropriate to use it here.

jsgilmore:
That is how I understood the purpose of tuple. @francis-liberty you'll note that both Emission and Immission have tuple fields. Although they are lists of objects, that represent the idea of a tuple in Storm. In addition to carrying tuples, emissions also contain all required meta data to emit them, but an emission can be any of: "emit", "sync", "log", "ack", "fail" or "error", as specified in the command field.

These emission types specify which actions the shell component will take in response to receiving an emission.

jsgilmore:
I could perhaps call an Immission a BoltMsg. Then emissions are messages sent from both spout and bolt non-JVM components to shell components, A SpoutMsg is a message sent from a ShellSpout to a non-JVM spout and a BoltMsg is a message sent from a ShellBolt to a non-JVM bolt.

I could take this pattern further and then call an emission a ShellMsg or a StormMsg, since it is a message sent from a non-JVM component to Storm or a Shell component. So this naming scheme would then be the destination of the message is its name.

xumingming:
haha, naming things is the most difficult part of programming. I think the new names you proposed are easier to understand than `Emission` and `Immission`.

francis-shuoch:
I think @jsgilmore 's naming convention is: w.r.t non-JVM component, "e-" means out, "i-" means in. So "emission" is to emit out of non-JVM component, "immission" is to emit in (or inject, actually) to non-JVM component. Is this more clear now?

jsgilmore:
So I've changed these names now to make them more understandable. But just for interest's sake: you don't have to play with "e-"s and "i-"s. They're just words for what happens from the non-JVM component's perspective.
From the free online dictionary:
e·mis·sion
n.
1. The act or an instance of emitting.
2. Something emitted.
http://www.thefreedictionary.com/Emission

Im`mis´sion
n.  1.  The act of immitting, or of sending or thrusting in; injection; - the correlative of emission.
http://www.thefreedictionary.com/immission

francis-shuoch:
Ok, thanks for these tips ;). Good job ~~

xumingming:
also, add some unit tests for this new feature.

jsgilmore:
My editor removes trailing whitespace. I thought this was required. The editor removed trailing whitespace in Config.java (of which there was a lot).

I couldn't really find shell component tests. backtype.storm.testing seems to have unit tests that make use of Storm's unit testing API for spouts and bolt. I couldn't really find any Storm component unit tests. How do you guys do unit tests in Storm? Where should my tests live? What do you use the run the tests (Maven?)?

Thanks for all the great feedback!

xumingming:
tests are placed in `storm-core/test/clj/backtype/storm`, tests are all writen in clojure, and run the tests with `lein test`

storm-core/src/jvm/backtype/storm/spout/ShellSpout.java
nathanmarz, line 38: Spout methods are never called concurrently by Storm. Any given spout object (remember the spout object is duplicated for each task that makes up the spout around the cluster) has all methods on it called from the same thread.

xumingming, line 38: thanks Nathan, then it's ok to have one SpoutMsg for all three message types. But personally I still think use a single SpoutMsg and change its property every time seems a little tricky.

jsgilmore:
I've had a look at the Storm tests. Specifically the shell component tests for ruby and python (multilang_test.clj). These tests also test the changes I made to Storm. They start up shell components and those shell components start to communicate. I didn't really add a feature to Storm, merely refactored code to be more modular.

storm-core/src/jvm/backtype/storm/spout/ShellSpout.java
jsgilmore, line 116: I've re-added this option.

xumingming:
Cool! I think we are almost there except that those pure whitespace-changes need to be removed(`project.clj`, `bootstrap.clj`, `Config.java`).

jsgilmore:
So I've reset the pure whitespace changes to project.clj and bootstrap.clj. Config.java was not a pure whitespace change; I added the STORM_MULTILANG_SERIALIZER item.

Just say if want me to re-add all the trailing whitespace in Config.java if that's what's required for the PR to be accepted :-)

xumingming:
For Config.java, I think the easiest way is to revert the current changes and add your new STORM_MULTIANG_SERIALIZER back.

jsgilmore:
Done

xumingming:
+1 for adding this PR into `0.9.1`

jsgilmore:
Sorry, I don't follow. Are you asking me to merge something more from upstream? I just now merged upstream master and there doesn't seem to be any changes and that's still on 0.9.0.

xumingming:
Oh, I think there is nothing more you need to do :) I am just voting to add your PR to the next release. Current release(`0.9.0`) only accept bugfix PR now.

PR #698, opened 2013-10-06
close #351. in storm.py: turn off stdout to avoid messing with JSON payload sent back to JAVA layer

francis-shuoch:
I write a python decorator to silence prints' output into stdout. Meanwhile sendMsgToParent() function doesn't get affected, so the communication to Java Layer is not affected, too. Use this wrapper on any custom python scripts containing prints without worrying those prints would mess up the tuple emits. I tested with storm-starter's WordCountTopology.

### DIFF
--- a/storm-core/src/multilang/py/storm.py
+++ b/storm-core/src/multilang/py/storm.py
@@ -55,9 +55,9 @@ def readTuple():
     return Tuple(cmd["id"], cmd["comp"], cmd["stream"], cmd["task"], cmd["tuple"])
 
 def sendMsgToParent(msg):
-    print json_encode(msg)
-    print "end"
-    sys.stdout.flush()
+    original_stdout.write(json_encode(msg) + '\n')
+    original_stdout.write("end\n")
+    original_stdout.flush()
 
 def sync():
     sendMsgToParent({'command':'sync'})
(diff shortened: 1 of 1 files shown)


xumingming:
So developers need to call this `nullify_stdout` method manually?

francis-shuoch:
Yes, as lbustelo pointed out in issue #351, custom python function containing prints can use this decorator to wrapper it, so user defined prints won't mess up the tuple emits from the bolts. Like this:

```
@nullify_stdout
def custom_fun:
    print "this won't come into python process's stdout to mess up emits."
```

sorry I didn't add an example like this in my PR.

francis-shuoch:
Another way is to assign sys.stdout = NullDevice() globally in the module, while sendMsgToParent() use original_stdout so it can write to stdout normally. But this way I don't know when to reassign sys.stdout back. Any thoughts here? I would like to follow it.

xumingming:
Would it be better if `storm.py` handles all the `silence prints` details, so it will be transparent to the developers?

francis-shuoch:
Then I think assigning sys.stdout = NullDevice() globally in the module while only sendMsgToParent() using original_stdout will suffice. What do you think?

xumingming:
About reassigning `sys.stdout` back, one way I can think of is wrapping all calls to `self.process(tuple)`,  `self.initialize(stormconf, context)`  etc with your `nullify_stdout`.

francis-shuoch:
I gave it some fresh thoughts today, now I think, since all prints will be read by storm's JVM-component, we must forbid any prints outside of `sendMsgToParent()` function. So, there is no need to reassign `sys.stdout` back. Does this make sense to you?

Or, alternatively, we can reassign `sys.stdout` back at the end of `run()` function, since this should be the last call of custom python script (see [splitsentence.py example](https://github.com/nathanmarz/storm-starter/blob/master/multilang/resources/splitsentence.py#L9) for reference).

PR #700, opened 2013-10-07
Makes JNI simpler on storm.

revans2:
This patch allows native libraries to be included in the topology jar like scripts for shell bolts and spouts.  The native library and its dependencies can be placed in either the resources directory or in resources/$(os.name)-$(os.arch)/

I have tested this manually, but I am not totally sure how to write a unit test for this.  It simply modifies the worker command line to include the appropriate directory.  After this is reviewed I am happy to also put up a pull request for the twiki describing how to use it.

### DIFF
--- a/storm-core/src/clj/backtype/storm/daemon/supervisor.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/supervisor.clj
@@ -400,12 +400,19 @@
       (FileUtils/moveDirectory (File. tmproot) (File. stormroot))
       ))
 
+(defn jlp [stormroot conf]
+  (let [resource-root (str stormroot "/" RESOURCES-SUBDIR)
+        os (clojure.string/replace (System/getProperty "os.name") #"\s+" "_")
+        arch (System/getProperty "os.arch")
+        arch-resource-root (str resource-root "/" os "-" arch)]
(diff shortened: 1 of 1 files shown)


storm-core/src/clj/backtype/storm/daemon/supervisor.clj
```diff
@@ -400,12 +400,19 @@
       (FileUtils/moveDirectory (File. tmproot) (File. stormroot))
       ))
 
+(defn jlp [stormroot conf]
+  (let [resource-root (str stormroot "/" RESOURCES-SUBDIR)
+        os (clojure.string/replace (System/getProperty "os.name") #"\s+" "_")
+        arch (System/getProperty "os.arch")
+        arch-resource-root (str resource-root "/" os "-" arch)]
+    (str arch-resource-root ":" resource-root ":" (conf JAVA-LIBRARY-PATH))))
```
d2r, line 408: Would it be nicer to use generic path separators here instead of `/`?

revans2:
I updated the code to not use a hard coded "/" separator.  All the unit tests still pass.

d2r:
I do realize that the code is surrounded by other instances of hard-coded `/`s :)  But at least we aren't adding more debt.

+1 (non-binding)

revans2:
This is superseded by https://github.com/apache/incubator-storm/pull/29

PR #701, opened 2013-10-08
Fixes shell input stream encoding bug

Virus-X:
There was an issue #119, but it was fixed only "by half": 
shell process output is read by Storm in some default encoding, which in my case was not UTF-8. Setting it explicitly fixed the problem.


xumingming:
+1

PR #702, opened 2013-10-09
fix the issue that tick tuple cannot work with system bolt

xumingming:
This patch fix #699 by not setup ticks for system components.


ptgoetz:
+1

(also assigning 0.9.0 milestone)

nathanmarz:
+1

ptgoetz:
Merging given that it came from a committer, has two binding +1 votes and has been open for 16 days.

PR #705, opened 2013-10-10
logviewer tries to determine log directory via logback configuration

strongh:
This introduces a new config variable, `logviewer.appender.name`, and modifies the logviewer process to locate the root log directory based on the the `FileAppender` named in the conf. Previously the logviewer assumed that the log directory was `{storm.home}/logs`.

If the named appender is not found or is not a `FileAppender` then the logviewer immediately throws a `RuntimeException`.

This issue is mentioned on the mailing list here: https://groups.google.com/d/msg/storm-user/IKRtIkqQfqc/OUDvyU6mxTMJ

### DIFF
--- a/storm-core/src/clj/backtype/storm/daemon/logviewer.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/logviewer.clj
@@ -31,14 +31,20 @@
     ))
 
 (defn log-root-dir
-  []
-  (let [appender (first (iterator-seq (.iteratorForAppenders (LoggerFactory/getLogger Logger/ROOT_LOGGER_NAME))))]
-    (if (and appender (instance? FileAppender appender))
-      (.getParent (File. (.getFile appender)))
-      (.getCanonicalPath (File. (System/getProperty "storm.home") "logs")))))
+  "Given an appender name, as configured, get the parent directory of the appender's log file.
 
-(defn log-page [file tail grep]
-  (let [path (.getCanonicalPath (File. (log-root-dir) file))
+Note that if anything goes wrong, this will throw an Error and exit."
+  [appender-name]
+  (let [appender (.getAppender (LoggerFactory/getLogger Logger/ROOT_LOGGER_NAME) appender-name)]
+    (if (and appender-name appender (instance? FileAppender appender))
+      (.getParent (File. (.getFile appender)))
+      (do
+        (throw
+         (Error. "Log viewer could not find configured appender, or the appender is not a FileAppender. Please check that the appender name configured in storm.yaml and cluster.xml agree."))
+        (System/exit 1)))))
+
+(defn log-page [file tail grep root-dir]
+  (let [path (.getCanonicalPath (File. root-dir file))
         tail (if tail
                (min 10485760 (Integer/parseInt tail))
                10240)
@@ -71,8 +77,8 @@
     ]))
 
 (defroutes log-routes
-  (GET "/log" [:as {cookies :cookies} & m]
-       (log-template (log-page (:file m) (:tail m) (:grep m))))
+  (GET "/log" [:as req & m]
+       (log-template (log-page (:file m) (:tail m) (:grep m) (:log-root req))))
   (GET "/loglevel" [:as {cookies :cookies} & m]
        (log-template (log-level-page (:name m) (:level m))))
   (route/resources "/")
@@ -82,10 +88,16 @@
   (handler/site log-routes)
  )
 
-(defn start-logviewer [port]
-  (run-jetty logapp {:port port}))
+(defn conf-middleware
+  "For passing the storm configuration with each request."
+  [app log-root]
+  (fn [req]
+    (app (assoc req :log-root log-root))))
+
+(defn start-logviewer [port conf]
+  (run-jetty (conf-middleware logapp conf) {:port port}))
 
 (defn -main []
-  (let [conf (read-storm-config)]
-    (start-logviewer (int (conf LOGVIEWER-PORT)))))
-
+  (let [conf (read-storm-config)
+        log-root (log-root-dir (conf LOGVIEWER-APPENDER-NAME))]
+    (start-logviewer (int (conf LOGVIEWER-PORT)) log-root)))
--- a/storm-core/src/clj/backtype/storm/daemon/logviewer.clj
+++ b/storm-core/src/clj/backtype/storm/daemon/logviewer.clj
@@ -38,10 +38,8 @@ Note that if anything goes wrong, this will throw an Error and exit."
   (let [appender (.getAppender (LoggerFactory/getLogger Logger/ROOT_LOGGER_NAME) appender-name)]
     (if (and appender-name appender (instance? FileAppender appender))
       (.getParent (File. (.getFile appender)))
-      (do
-        (throw
-         (Error. "Log viewer could not find configured appender, or the appender is not a FileAppender. Please check that the appender name configured in storm.yaml and cluster.xml agree."))
-        (System/exit 1)))))
+      (throw
+       (RuntimeException. "Log viewer could not find configured appender, or the appender is not a FileAppender. Please check that the appender name configured in storm and logback agree.")))))
 
 (defn log-page [file tail grep root-dir]
   (let [path (.getCanonicalPath (File. root-dir file))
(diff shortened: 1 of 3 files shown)


strongh:
#432 is relevant but this PR isn't necessarily a complete fix.

storm-core/src/clj/backtype/storm/daemon/logviewer.clj
```diff
@@ -25,8 +29,17 @@
       (.toString output))
     ))
 
+(defn log-root-dir
+  []
+  (let [appender (first (iterator-seq (.iteratorForAppenders (LoggerFactory/getLogger Logger/ROOT_LOGGER_NAME))))]
+    (if (and appender (instance? FileAppender appender))
+      (string/join
+       "/" (butlast
+            (string/split (.getFile appender) #"/")))
```
d2r: Probably want to strip all of the trailing path separators while we're at it.
Could also try something like `(.getCanonicalPath (File. (.getFile appender)))`?

storm-core/src/clj/backtype/storm/daemon/logviewer.clj
```diff
@@ -25,8 +29,17 @@
...
+      (str (System/getProperty "storm.home") "/logs/"))))
+
 (defn log-page [file tail grep]
-  (let [path (str (System/getProperty "storm.home") "/logs/" file)
+  (let [path (str (log-root-dir) "/" file)
```
d2r: Could we use `File/pathSeparator` instead of `/` in general?
Or `(.getCanonicalPath (clojure.java.io/file (log-root-dir) file))`

storm-core/src/clj/backtype/storm/daemon/logviewer.clj
strongh: Sounds good to me.

storm-core/src/clj/backtype/storm/daemon/logviewer.clj
```diff
@@ -25,8 +30,15 @@
       (.toString output))
     ))
 
+(defn log-root-dir
+  []
+  (let [appender (first (iterator-seq (.iteratorForAppenders (LoggerFactory/getLogger Logger/ROOT_LOGGER_NAME))))]
+    (if (and appender (instance? FileAppender appender))
+      (.getParent (File. (.getFile appender)))
+      (.getCanonicalPath (File. (System/getProperty "storm.home") "logs")))))
```
d2r: So if I read this correctly, this expects that the first appender defined is a `FileAppender`, and it assumes that this appender writes to the directory that all of the interesting logs are written to as well.

I wonder if we should consider standardizing the appender name and search for that explicitly?  Otherwise we impose a requirement on the structure of the logback configuration.

strongh: Yeah, there are definitely assumptions made. My goal here was just to allow strictly more use cases than previously supported, including mine and those mentioned in the mailing list.

I'm not sure what the best/conventional way to handle this without those assumptions. Standardizing the appender name is one option. I also had thought about configuring the the appender name as a logviewer-specific variable in the YAML. It would mean that some folks would have to configure the appender name twice (once in the YAML and once in the logback XML) which is annoying, but maybe worth the flexibility for the few who need it.

d2r: You're right.  So the options so far:
1.  Use a new convention that requires the needed appender first in the logback config
2.  Use a new convention that says the needed appender must have a certain name ("A1"?)
3.  Add a config that stores the name of the needed appender, and make sure that appender has the same name.

If we had reasonable defaults, option 3 does not seem so bad.  A reasonable default would be "A1," so even current installations would not need to change anything to have logs and the logviewer continue to work.

In theory, (following your suggestion) someone may want to launch different logviewers to read different logs.  In that case, there could be a common logback config with different appenders, and each logviewer would need its own appender to look for.  Option 3 could also support this use case.  (E.g., a logviewer for "METRICS" only) This is a use case that makes it necessary instead of redundant to configure the appender name in both the storm and logback configs.

strongh: Makes sense. Sounds to me like option 3 is the way to go. Allows for flexible configuration and works by default. To be clear:
- add a new config variable `logviewer.appender.name` whose value defaults to `"A1"`, and
- logviewer will serve logs out of the parent directory of  `logviewer.appender.name`'s log file.

The only minor part that's still not clear to me is the behavior if the configured appender is not found or is not a `FileAppender`. Maybe just log a warning? Question is whether to use a fallback or throw an exception.

d2r: Good question.  If we just log a warning and go on, then what will the logviewer go on doing?  It seems like this would be a fatal error requiring an exception (with detail of why it failed).

strongh:
@d2r took a pass at the changes we discussed, let me know what you think.

storm-core/src/clj/backtype/storm/daemon/logviewer.clj
```diff
@@ -25,8 +30,21 @@
...
+  "Given an appender name, as configured, get the parent directory of the appender's log file.
+
+Note that if anything goes wrong, this will throw an Error and exit."
+  [appender-name]
+  (let [appender (.getAppender (LoggerFactory/getLogger Logger/ROOT_LOGGER_NAME) appender-name)]
+    (if (and appender-name appender (instance? FileAppender appender))
+      (.getParent (File. (.getFile appender)))
+      (do
+        (throw
+         (Error. "Log viewer could not find configured appender, or the appender is not a FileAppender. Please check that the appender name configured in storm.yaml and cluster.xml agree."))
```
d2r: I think we could throw a `RuntimeException` here.  Errors are reserved for things like bad JVM state.

It is great the message refers to the storm configuration and the logback configuration.  It might be better to leave off the explicit file names though.

storm-core/src/clj/backtype/storm/daemon/logviewer.clj
```diff
@@ -25,8 +30,21 @@
...
+        (System/exit 1)))))
```
d2r: We are throwing an exception just before this, so we will never get here.

storm-core/src/clj/backtype/storm/daemon/logviewer.clj
```diff
@@ -70,10 +88,16 @@
...
+  "For passing the storm configuration with each request."
+  [app log-root]
+  (fn [req]
+    (app (assoc req :log-root log-root))))
 
-(defn -main []
-  (let [conf (read-storm-config)]
-    (start-logviewer (int (conf LOGVIEWER-PORT)))))
+(defn start-logviewer [port conf]
+  (run-jetty (conf-middleware logapp conf) {:port port}))
```
d2r: The `conf` parameter is really only `log-root`, right?

strongh: Good point.

storm-core/src/clj/backtype/storm/daemon/logviewer.clj
strongh: I've confirmed that you are correct. Not sure why I thought it was necessary.

d2r:
OK, I tested the default installation, changing the appender name & config, moving the appender definition below "ACCESS" in the XML, and changing the `file` value to a different directory.  The logviewer was able to find the correct logs in each case.

Also tested that an invalid appender name in storm config throws.

+1 (non-binding)

We will want @nathanmarz, other committers to take a look.

strongh:
@d2r thanks for the feedback and for trying it out. I've updated the PR description to better describe how the behavior ended up.

ptgoetz:
I'm going to tag this for the 0.9.0 release.

One thing I think we need is documentation on configuration, etc. perhaps as a readme.md in the module directory... How to configure the log viewer, what it needs,...

We regularly modify the log config for centralized logging, etc. and most sysadmins are going to see logs in something like /var/log

ptgoetz:
+1

xiaokang:
+1

ptgoetz:
Bump. 

Looking for additional reviews from committers.

xumingming:
+1

PR #706, opened 2013-10-10
Do the worker HB timeout check when HB's are updated

d2r:
Instead of asking zookeeper for the latest heartbeat data for all topologies and then checking which heartbeats have timed out, combine the timeout check with the heartbeat update.

This makes nimbus more tolerant of a slow zookeeper server, in which case the updating of heartbeats can take so long that by the time the nimbus actually examines the new heartbeats for timeouts, most of the earliest topologies have complete timeouts of executors.


d2r:
This pull request is initially intended to start discussion.

Current unit tests pass with this change.

d2r:
This could mitigate a possible root cause for #689.

brndnmtthws:
:+1: 

Good stuff.

ptgoetz:
+1
(tested in distributed mode with 5-node cluster)

nathanmarz:
+1

xumingming:
+1

PR #710, opened 2013-10-16
Fix sys stats cookie being lost from UI requests

d2r:
Undo part of a change introduced with 7694401ddf8bc8c988dd70414f1aae8d1540e02b

This fixes a possible regression from 0.9.0 introduced in 0.9.0-rc1

### DIFF
--- a/storm-core/src/clj/backtype/storm/ui/core.clj
+++ b/storm-core/src/clj/backtype/storm/ui/core.clj
@@ -803,7 +803,7 @@
         ))))
 
 (def app
-  (-> #'main-routes
+  (-> main-routes
       (wrap-reload '[backtype.storm.ui.core])
       catch-errors))
 
--- a/storm-core/src/clj/backtype/storm/ui/core.clj
+++ b/storm-core/src/clj/backtype/storm/ui/core.clj
@@ -803,9 +803,9 @@
         ))))
 
 (def app
-  (-> main-routes
-      (wrap-reload '[backtype.storm.ui.core])
-      catch-errors))
+  (handler/site (-> main-routes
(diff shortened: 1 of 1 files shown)


ptgoetz:
Added to 0.9.0 milestone.

ptgoetz:
+1

xumingming:
+1

d2r:
The previous commit may only fix the issue by accident as a special case.

Cookies need to be wrapped properly so that they can be bound properly in the destructuring of the request.  I noticed this when I applied the same fix to a version of the ui running with servlet-request parameters, and in that test the :cookies were not bound correctly.

d2r:
New patch properly wraps cookies as before.  I also verified the code reloads when the source file changes via the repl.

ptgoetz:
Bump. Looking for more community review.

revans2:
I am not an expert on this, but all of the compojure tutorials I have seen call compojure.handler/site on the routes.
Even the log-viewer app is doing it.  I am +1 on the change.

Full disclosure: I work with Derek

d2r:
To clarify what happens:

compojure.handler/site -> ring.middleware.session/wrap-session -> ring.middleware.cookies/wrap-cookies

compojure.handler/site does other nice things, but wrapping the cookies is what we need to restore the functionality.  I do not think the rest of compojure.handler/site hurts anything.  Also, that is what was in core.clj before the auto-reload patch anyway.

d2r:
Closed on accident, reopening

anfeng:
+1

ptgoetz:
Thanks @anfeng. Merging for rc3.

PR #715, opened 2013-10-21
Ability to parse multi value -c arguments on the commandline

enachb:
Hi!

The following patch enables lists of values to be parsed on the CLI via the `-c` argument.
This way one can configure Nimbus and Supervisors almost entirely from the CLI.
It is fully backwards compatible and only provides better parsing logic to distinguish keys from lists of values.

Sample:

```
bin/storm-mesos nimbus -c java.library.path=native:/usr/local/lib,mesos.master.url=master1:5050,mesos.executor.uri=http://mydomain.com/storm-mesos-0.8.3-wip3.tgz,nimbus.host=myMaster,storm.local.dir=../storm-local,ui.port=7070,storm.zookeeper.servers=zk0,zk1,zk2
```

We use it to run Storm on Mesos, but the patch is generic and benefits regular Storm CLI parsing.
I do have tests, but didn't see a standard way to trigger tests for utility functions like this.
Thanks!
-Erich

### DIFF
--- a/storm-core/src/jvm/backtype/storm/utils/Utils.java
+++ b/storm-core/src/jvm/backtype/storm/utils/Utils.java
@@ -136,17 +136,44 @@ public class Utils {
     public static Map readDefaultConfig() {
         return findAndReadConfigFile("defaults.yaml", true);
     }
-    
+
     public static Map readCommandLineOpts() {
         Map ret = new HashMap();
         String commandOptions = System.getProperty("storm.options");
-        if(commandOptions != null) {
+        if (commandOptions != null) {
             commandOptions = commandOptions.replaceAll("%%%%", " ");
+
             String[] configs = commandOptions.split(",");
-            for (String config : configs) {
-                String[] options = config.split("=");
-                if (options.length == 2) {
+            String lastArgKey = null;
+            for (int i = 0; i < configs.length; i++) {
+                if (configs[i].contains("=")) {
+                    // we have a key value pair -> split it
+                    String[] options = configs[i].split("=");
                     ret.put(options[0], options[1]);
+                    lastArgKey = options[0];
+                } else {
+                    // list value, please add to previous full one
+                    ret.put(lastArgKey, ret.get(lastArgKey) + "," + configs[i]);
+                }
+            }
+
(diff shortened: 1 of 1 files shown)


d2r:
This is cool.  Can it handle Maps, as with `isolation.scheduler.machines`?

enachb:
It currently doesn't support Maps, but probably could. 
Colons are fairly common in strings (e.g. port numbers in URLs), so one would need to check for double quotes before the colon to decide if it is a map or not.

revans2:
For testing you could easily split the core of this out into a function that takes a single string argument, and then have the original call this new function with System.getProperty("storm.options").

When creating a list the code converts numeric strings into longs, but that does not happen with single value arguments.  It would be good if it were consistent.  It would also be nice to support boolean values, we have had issues where stuff does not work as expected because "false" does not always evaluate to false.

enachb:
I do have tests which use the `storm.options` property for testing. I was just not sure where to include them.
A pointer would be helpful.

Good points:
- I will add single longs to the conversion logic and
- will also add single and list boolean values

storm-core/src/jvm/backtype/storm/utils/Utils.java
```diff
@@ -136,17 +137,42 @@ public static Map findAndReadConfigFile(String name) {
...
+                    ret.put(lastArgKey, ret.get(lastArgKey) + "," + configs[i]);
+                }
+            }
+
+            // If we have a list argument make it an actual list of the right type
+            for (Object key : ret.keySet()) {
+                String[] options = ret.get(key).toString().split(",");
+                if (options.length > 1) {
+                    // make a list out of it otherwise leave it as is
+                    if (StringUtils.isNumeric(options[0])) {
```
anfeng, line 166: We need to be consistent for data type support. If we distinguish number vs string for lists, we should do the same for non-lists.

enachb:
Closing this pull request as it was meant to support running Storm on Mesos.

I have been able to accomplish the same by embedding an HTTP server into Nimbus. The Supervisors upon launch used it to pull their configuration down:
https://github.com/mesosphere/storm-mesos/commit/41fbf0ac41487e5dda019cb4946bb5112f9d0985

PR #719, opened 2013-10-23
Adding error command handling to ShellSpout.

vinodc:
The ShellBolt has an equivalent case for handling "error" commands so I thought it would be appropriate for the ShellSpout to have one too.

### DIFF
--- a/storm-core/src/jvm/backtype/storm/spout/ShellSpout.java
+++ b/storm-core/src/jvm/backtype/storm/spout/ShellSpout.java
@@ -88,6 +88,10 @@ public class ShellSpout implements ISpout {
                 } else if (command.equals("log")) {
                     String msg = (String) action.get("msg");
                     LOG.info("Shell msg: " + msg);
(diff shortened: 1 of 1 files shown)


storm-core/src/jvm/backtype/storm/spout/ShellSpout.java
```diff
@@ -88,6 +88,10 @@ private void querySubprocess(Object query) {
                 } else if (command.equals("log")) {
                     String msg = (String) action.get("msg");
                     LOG.info("Shell msg: " + msg);
+                } else if (command.equals("error")) {
+                    String msg = (String) action.get("msg");
+                    _collector.reportError(new Exception("Shell Process Exception: " + msg));
+                    return;
```
vinodc, line 94: If an error is received from a spout, we should probably `return` to allow the subprocess to be queried again. Otherwise we would be waiting on further messages from the subprocess which is unlikely considering we received an error. That could result in the process waiting indefinitely.

PR #722, opened 2013-10-26
Support for running on Windows

davidlao2k:
This is a minimal set of changes to support running storm 0.9.0 on Windows. Please discard the previous pull request as a few of the changes were done on older branch and no longer needed.

### DIFF
--- a/src/clj/backtype/storm/daemon/supervisor.clj
+++ b/src/clj/backtype/storm/daemon/supervisor.clj
@@ -317,7 +317,14 @@
         (when-not (assigned-storm-ids storm-id)
           (log-message "Removing code for storm id "
                        storm-id)
-          (rmr (supervisor-stormdist-root conf storm-id))
+          (try
+            (rmr (supervisor-stormdist-root conf storm-id))
+            (catch java.io.IOException e
+              ;; Handles a race condition resulting from topology kill where this sync call
+              ;; can come in before the worker processes are killed. On Linux/MacOs this works fine
+              ;; but on Windows rmr will throw "Unable to delete file" IOException because the resources
+              ;; are still being used. After killing the next sync call should clean things up nicely.
+              (log-message (.getMessage e))))
           ))
       (.add processes-event-manager sync-processes)
       )))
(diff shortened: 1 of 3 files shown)


src/clj/backtype/storm/daemon/supervisor.clj
```diff
@@ -317,7 +317,14 @@
         (when-not (assigned-storm-ids storm-id)
           (log-message "Removing code for storm id "
                        storm-id)
-          (rmr (supervisor-stormdist-root conf storm-id))
+          (try
+            (rmr (supervisor-stormdist-root conf storm-id))
+            (catch java.io.IOException e
+              ;; Handles a race condition resulting from topology kill where this sync call
```
nathanmarz, line 323: There's legitimate cases where an IOException is thrown here and the rmr did not go through. This is a case where we should check if we're running in Windows and take different behavior as opposed to Mac/Linux (which should rethrow the exception)

nathanmarz:
Everything looks good except for that one comment I made. Once that gets cleaned up I'm +1

oppokui:
It is wonderful!
By the way, do you enhance storm python script? 
I changed 0.9.0-rc2 a little, share diff in https://app.box.com/s/8k2czivqfzxyz3ixksnp. 
After it, I can use ">call python storm nimbus" to start nimbus in Windows OS.

src/clj/backtype/storm/daemon/supervisor.clj
davidlao2k, line 323: Updated. Thanks for review.

ChitturiPadma:
Hi,
I have taken storm-0.8.2, replaced .sh files and storm python script with .cmd files that exist in bin folder of https://github.com/davidlao2k/storm-windows. sbin is also merged into storm-0.8.2 which contains .dlls of jmq and omq dependencies. Upon running bin>storm nimbus from the cmd line, nimbus starts but nimbus.log is not seen in \logs folder. Is it the correct way to launch nimbus and supervisors from .cmd scripts or need to run the storm python script?

ChitturiPadma:
Hi Folks, 
Storm on Windows is successful !! Able to submit topology in distributed environment using maven. Traditional way of compiling the source code using javac and submission of topology is throwing errors. Need to test submitting topologies using different ways (using storm jars, leiningen, maven etc).

ChitturiPadma:
when i kill the topology, the supervisors seem to abort. This behavior is strange !!! Need to look into this

ptgoetz:
This was merged into the 0.9.0 branch, not master, so it didn't get included in any of the 0.9.0 releases.

@davidlao2k do you want to resubmit a pull request (preferably to the apache/incubator-storm repo) so we can get this integrated, or do you want me to handle it?

Taylor

davidlao2k:
Taylor, please go ahead and handle the resubmit. Thanks for help.

PR #725, opened 2013-10-28
Retry LocalState snapshot on wrapped io exceptions

d2r:
`Utils#deserialize` now throws IOExceptions wrapped in RuntimeExceptions, so we need to account for that.

Tested OK with word_count topology from storm-starter.


d2r:
Closed on accident, reopening

d2r:
Fixed [here](https://github.com/apache/storm/blob/ced622433eaff81d7039d384c3acf5de595b0a70/storm-core/src/jvm/backtype/storm/utils/LocalState.java#L58)

PR #713, opened 2013-10-18
Fixed negative netty sleep values.

brndnmtthws:
For some context, it seems that retries.get() occasionally returns negative values.

PR #726, opened 2013-10-30
Ensure we don't overflow the backoff value.

brndnmtthws:
The first attempt to fix this (213102b36f890) did not correctly address
the issue.  The 32 bit signed integer frequently overflows, resulting in
a bad value for Random.nextInt().

See previous pull request @ https://github.com/nathanmarz/storm/pull/713

Here's the exception:

```
2013-10-30 16:30:04 b.s.m.n.Client [INFO] Reconnect ... [26]
2013-10-30 16:30:04 b.s.m.n.Client [INFO] Reconnect ... [29]
2013-10-30 16:30:04 b.s.m.n.Client [INFO] Reconnect ... [27]
2013-10-30 16:30:04 b.s.m.n.Client [INFO] Reconnect ... [30]
2013-10-30 16:30:04 STDIO [ERROR] Oct 30, 2013 4:30:04 PM org.jboss.netty.channel.DefaultChannelPipeline
WARNING: An exception was thrown by a user handler while handling an exception event ([id: 0x27388d8c] EXCEPTION: java.net.ConnectException: Connection refused: i-0e040e75/10.58.82.253:31002)
java.lang.IllegalArgumentException: n must be positive
    at java.util.Random.nextInt(Random.java:300)
    at backtype.storm.messaging.netty.Client.getSleepTimeMs(Client.java:97)
    at backtype.storm.messaging.netty.Client.reconnect(Client.java:78)
    at backtype.storm.messaging.netty.StormClientHandler.exceptionCaught(StormClientHandler.java:108)
    at org.jboss.netty.handler.codec.frame.FrameDecoder.exceptionCaught(FrameDecoder.java:377)
    at org.jboss.netty.channel.Channels.fireExceptionCaught(Channels.java:525)
    at org.jboss.netty.channel.socket.nio.NioClientBoss.processSelectedKeys(NioClientBoss.java:110)
    at org.jboss.netty.channel.socket.nio.NioClientBoss.process(NioClientBoss.java:79)
    at org.jboss.netty.channel.socket.nio.AbstractNioSelector.run(AbstractNioSelector.java:312)
    at org.jboss.netty.channel.socket.nio.NioClientBoss.run(NioClientBoss.java:42)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
    at java.lang.Thread.run(Thread.java:744)
2013-10-30 16:30:05 b.s.m.n.Client [INFO] Reconnect ... [29]
2013-10-30 16:30:06 b.s.m.n.Client [INFO] Reconnect ... [28]
```

### DIFF
--- a/storm-netty/src/jvm/backtype/storm/messaging/netty/Client.java
+++ b/storm-netty/src/jvm/backtype/storm/messaging/netty/Client.java
@@ -23,14 +23,14 @@ import backtype.storm.utils.Utils;
 
 class Client implements IConnection {
     private static final Logger LOG = LoggerFactory.getLogger(Client.class);
-    private final int max_retries; 
-    private final int base_sleep_ms; 
-    private final int max_sleep_ms; 
+    private final int max_retries;
+    private final int base_sleep_ms;
+    private final int max_sleep_ms;
     private LinkedBlockingQueue<Object> message_queue; //entry should either be TaskMessage or ControlMessage
     private AtomicReference<Channel> channelRef;
     private final ClientBootstrap bootstrap;
     private InetSocketAddress remote_addr;
-    private AtomicInteger retries; 
+    private AtomicInteger retries;
     private final Random random = new Random();
     private final ChannelFactory factory;
     private final int buffer_size;
@@ -38,14 +38,14 @@ class Client implements IConnection {
 
     @SuppressWarnings("rawtypes")
     Client(Map storm_conf, String host, int port) {
-        message_queue = new LinkedBlockingQueue<Object>(); 
+        message_queue = new LinkedBlockingQueue<Object>();
         retries = new AtomicInteger(0);
         channelRef = new AtomicReference<Channel>(null);
         being_closed = new AtomicBoolean(false);
 
-        // Configure 
+        // Configure
         buffer_size = Utils.getInt(storm_conf.get(Config.STORM_MESSAGING_NETTY_BUFFER_SIZE));
-        max_retries = Utils.getInt(storm_conf.get(Config.STORM_MESSAGING_NETTY_MAX_RETRIES));
+        max_retries = Math.min(30, Utils.getInt(storm_conf.get(Config.STORM_MESSAGING_NETTY_MAX_RETRIES)));
         base_sleep_ms = Utils.getInt(storm_conf.get(Config.STORM_MESSAGING_NETTY_MIN_SLEEP_MS));
         max_sleep_ms = Utils.getInt(storm_conf.get(Config.STORM_MESSAGING_NETTY_MAX_SLEEP_MS));
         int maxWorkers = Utils.getInt(storm_conf.get(Config.STORM_MESSAGING_NETTY_CLIENT_WORKER_THREADS));
@@ -74,9 +74,9 @@ class Client implements IConnection {
     void reconnect() {
         try {
             int tried_count = retries.incrementAndGet();
-            if (tried_count < max_retries) {
+            if (tried_count <= max_retries) {
                 Thread.sleep(getSleepTimeMs());
-                LOG.info("Reconnect ... [{}]", tried_count);   
+                LOG.info("Reconnect ... [{}]", tried_count);
                 bootstrap.connect(remote_addr);
                 LOG.debug("connection started...");
             } else {
@@ -85,7 +85,7 @@ class Client implements IConnection {
             }
         } catch (InterruptedException e) {
             LOG.warn("connection failed", e);
-        } 
+        }
     }
 
     /**
@@ -93,19 +93,17 @@ class Client implements IConnection {
      */
     private int getSleepTimeMs()
     {
-        int backoff = 1 << Math.max(1, retries.get());
+        int backoff = 1 << retries.get();
         int sleepMs = base_sleep_ms * Math.max(1, random.nextInt(backoff));
         if ( sleepMs > max_sleep_ms )
             sleepMs = max_sleep_ms;
-        if ( sleepMs < base_sleep_ms )
-          sleepMs = base_sleep_ms;
         return sleepMs;
     }
 
     /**
-     * Enqueue a task message to be sent to server 
+     * Enqueue a task message to be sent to server
      */
-    public void send(int task, byte[] message) {        
+    public void send(int task, byte[] message) {
         //throw exception if the client is being closed
         if (being_closed.get()) {
             throw new RuntimeException("Client is being closed, and does not take requests any more");
@@ -128,43 +126,43 @@ class Client implements IConnection {
         MessageBatch batch = new MessageBatch(buffer_size);
         Object msg = message_queue.take();
         batch.add(msg);
-        
+
         //we will discard any message after CLOSE
-        if (msg==ControlMessage.CLOSE_MESSAGE) 
+        if (msg==ControlMessage.CLOSE_MESSAGE)
             return batch;
-        
+
         while (!batch.isFull()) {
             //peek the next message
             msg = message_queue.peek();
             //no more messages
             if (msg == null) break;
-            
+
             //we will discard any message after CLOSE
             if (msg==ControlMessage.CLOSE_MESSAGE) {
                 message_queue.take();
                 batch.add(msg);
                 break;
             }
-            
+
             //try to add this msg into batch
             if (!batch.tryAdd((TaskMessage) msg))
                 break;
-            
+
             //remove this message
             message_queue.take();
         }
 
         return batch;
     }
-    
+
     /**
      * gracefully close this client.
-     * 
+     *
      * We will send all existing requests, and then invoke close_n_release() method
      */
     public synchronized void close() {
-        if (!being_closed.get()) {  
-            //enqueue a CLOSE message so that shutdown() will be invoked 
+        if (!being_closed.get()) {
+            //enqueue a CLOSE message so that shutdown() will be invoked
             try {
                 message_queue.put(ControlMessage.CLOSE_MESSAGE);
                 being_closed.set(true);
@@ -178,10 +176,10 @@ class Client implements IConnection {
      * close_n_release() is invoked after all messages have been sent.
      */
     void  close_n_release() {
-        if (channelRef.get() != null) 
+        if (channelRef.get() != null)
             channelRef.get().close().awaitUninterruptibly();
 
-        //we need to release resources 
+        //we need to release resources
         new Thread(new Runnable() {
             @Override
             public void run() {
@@ -194,10 +192,10 @@ class Client implements IConnection {
     }
 
     void setChannel(Channel channel) {
-        channelRef.set(channel); 
-        //reset retries   
+        channelRef.set(channel);
+        //reset retries
         if (channel != null)
-            retries.set(0); 
+            retries.set(0);
     }
 
 }
--- a/conf/defaults.yaml
+++ b/conf/defaults.yaml
@@ -86,7 +86,7 @@ zmq.hwm: 0
 storm.messaging.netty.server_worker_threads: 1
 storm.messaging.netty.client_worker_threads: 1
 storm.messaging.netty.buffer_size: 5242880 #5MB buffer
-storm.messaging.netty.max_retries: 100
+storm.messaging.netty.max_retries: 30
 storm.messaging.netty.max_wait_ms: 1000
 storm.messaging.netty.min_wait_ms: 100
 


storm-netty/src/jvm/backtype/storm/messaging/netty/Client.java
```diff
@@ -93,8 +93,8 @@ void reconnect() {
      */
     private int getSleepTimeMs()
     {
-        int backoff = 1 << Math.max(1, retries.get());
-        int sleepMs = base_sleep_ms * Math.max(1, random.nextInt(backoff));
+        int backoff = 1 << Math.min(31, retries.get());
```
d2r: - This implies a ceiling of 31 retries for storm.messaging.netty.max_retries.
- According to [this](https://github.com/brndnmtthws/storm/blob/88d23962f5f5f10505cefe2ef300a8f19b353bba/storm-netty/src/jvm/backtype/storm/messaging/netty/Client.java#L77), we can retry at most storm.messaging.netty.max_retries - 1 times, since we increment on the first re-try.  So the config is a bit of a misnomer.
- Newer versions of curator (starting with 1.0.15 I think) have a BoundedExponentialBackoffRetry.  Since we are using an older curator, we sub-classed ExponentialBackoffRetry to add the bounds [here](https://github.com/nathanmarz/storm/blob/master/storm-core/src/jvm/backtype/storm/utils/Utils.java#L313).  This was originally added so that workers coming up after a ZK failure would not overwhelm the ZK cluster. I wonder if we could re-use the `Util.BoundedExponentialBackoffRetry` here, since it is tested, or possibly just look at the implementation in the curator code?

brndnmtthws: It's a 32 bit signed integer, so any value of `retries` exceeding 31 will cause the value to overflow.

brndnmtthws: To your other point about curator: can we just upgrade Curator to 1.0.22 (I think that's the stable 1.0.x version)?  I'm not sure there's much merit for duplicating efforts.

d2r: > It's a 32 bit signed integer, so any value of retries exceeding 31 will cause the value to overflow.

Yeah you're right on—I was raising a question around the implementation of the back-off in general.

(EDIT to add quote)

brndnmtthws: Regarding the implementation, I'm in favour of updating curator and using that.

d2r: > To your other point about curator: can we just upgrade Curator to 1.0.22 (I think that's the stable 1.0.x version)? I'm not sure there's much merit for duplicating efforts.

Upgrading curator has been discussed numerous times, and that would probably best be started as a separate issue.  I think an upgrade was even attempted in the past, but reverted.

The place to go for ZK/Curator upgrades is now #480.

brndnmtthws: Got it.

In the mean time, this should be merged.

brndnmtthws:
@ptgoetz Can you take a look? We've been using this for a few weeks now.

ptgoetz:
This will fix the problem of negative back off values, but I think there are some issues with the logic.

The backoff value increases exponentially, but that is largely negated by the `random.nextInt()` call. As it stands, the logic in this pull request more or less uses a random value between `min_sleep_ms` and `max_sleep_ms` for the amount of time to sleep between retries.

If you extract the logic and run it an `main()` method, you'll quickly see what I mean:

``` java
    public static void main(String[] args) throws Exception {
        AtomicInteger retries = new AtomicInteger(0);
        Random random = new Random();
        int max_retries = 100;
        int base_sleep_ms = 1000;
        int max_sleep_ms = 5000;

        while(true){
            int tried_count = retries.incrementAndGet();
            if (tried_count < max_retries) {
                int backoff = 1 << Math.min(30, retries.get());
                System.out.println("Backoff value: " + backoff);
                int sleepMs = base_sleep_ms * random.nextInt(backoff);
                if ( sleepMs > max_sleep_ms )
                    sleepMs = max_sleep_ms;
                if ( sleepMs < base_sleep_ms )
                    sleepMs = base_sleep_ms;

                System.out.println("Reconnect #" + tried_count + " Sleep time: " + sleepMs);

            } else {
                System.out.println("Remote address is not reachable. We will close this client.");
                break;
            }
        }
```

I would eliminate the random factor and use `backoff * some_value` to the sleep time actually increases progressively.

Regardless, it should be fixed. I'll add it to the 0.9.0 milestone. Unfortunately it won't make it into rc3 since that's already been cut, but I don't have a problem with releasing an rc4 on the heels of rc3.

brndnmtthws:
To be honest, I don't quite understand the intention of the original code.  It shouldn't be throwing exceptions though.

d2r:
Logic looks similar to BoundedExponentialBackoffRetry from Curator.  See https://github.com/nathanmarz/storm/pull/726#discussion_r7318475

ptgoetz:
@brndnmtthws Do you have any sample code/config that can consistently reproduce this issue? I've not seen this issue "in the wild".

@d2r Per my previous comment, if the `BoundedExponentialBackoffRetry` implementation from Curator uses the same logic, then I'm not sure we want to follow suit. I'm not sure adding the random component is the right thing to do. I would think we would want the back-off functionality to be predictable.

Even if we just eliminate the possibility of negative values and the resulting error, I'd like to target this for 0.9.0.

d2r:
@ptgoetz, The back-off is intended to protect a resource from getting slammed with requests coming in at the same time from a lot of clients.  If the back-off were not randomized, then after an outage the resource would get slammed at exponentially increasing intervals, as each client would issue request at around the same predictable time.  The logic has successfully allowed recovery of ZooKeeper after an outage, when otherwise synchronized requests were overwhelming.

ptgoetz:
@d2r Fair enough. The important part is to eliminate the error resulting from negative values.

+1 for 0.9.0

brndnmtthws:
@ptgoetz It would occur pretty frequently under normal operation (without this patch).  The cause seems to be that was that our workers would take longer to start up than perhaps normally expected (we're running Storm on Mesos, btw) so there was greater likelihood of hitting this.

nathanmarz:
@brndnmtthws: Can you submit a contributor agreement to Apache? http://www.apache.org/licenses/icla.txt We need that before we can merge this in.

nathanmarz:
I'm not that familiar with the storm-netty code, so I'd like to see +1's from @anfeng and the others who worked on it.

brndnmtthws:
I'm already an Apache contributor for the Mesos project.

anfeng:
-1. I don't like we use a hard coded value 30. Instead, we should set default.yaml correctly. 

Currently, we have 100. It should be 31 for 32-bit OS or 63 for 64-bit OS.
    storm.messaging.netty.max_retries: 100

brndnmtthws:
An `int` is 32 bits regardless of whether you're using x86 or x86-64.

anfeng:
Shouldn't we change the default as below?

storm.messaging.netty.max_retries: 31

brndnmtthws:
I don't see how that fixes the problem.  And 31 is 1 too many bits.

anfeng:
31 is okay duw to the logic in reconnect().

void reconnect() {
        try {
            int tried_count = retries.incrementAndGet();
            if (tried_count < max_retries) {
                Thread.sleep(getSleepTimeMs());
                LOG.info("Reconnect ... [{}]", tried_count);
                bootstrap.connect(remote_addr);
                LOG.debug("connection started...");
            } else {
                LOG.warn("Remote address is not reachable. We will close
this client.");
                close();
            }
        } catch (InterruptedException e) {
            LOG.warn("connection failed", e);
        } 
    }

From:  Brenden Matthews notifications@github.com
Reply-To:  nathanmarz/storm
<reply+i-21839840-8a4217e689f10a2241b0ed54b848f671f8516523-2708344@reply.git
hub.com>
Date:  Tuesday, December 3, 2013 2:26 PM
To:  nathanmarz/storm storm@noreply.github.com
Cc:  Andy Feng andy.feng@gmail.com
Subject:  Re: [storm] Ensure we don't overflow the backoff value. (#726)

I don't see how that fixes the problem.  And 31 is 1 too many bits.

‹
Reply to this email directly or view it on GitHub
https://github.com/nathanmarz/storm/pull/726#issuecomment-29758103 .

brndnmtthws:
I don't think that's right:

```
python
Python 2.7.6 (default, Nov 19 2013, 14:05:42) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import ctypes
>>> ctypes.c_int32(1 << 30)
c_int(1073741824)
>>> ctypes.c_int32(1 << 31)
c_int(-2147483648)
>>> 
```

HeartSaVioR:
I misunderstood the problem. (removed previous comment)

@anfeng you're right.
if statement evaluates to 'false' when tried_count == max_retries, so if max_retries is 31, maximum sleep seconds are 1 << 30.

HeartSaVioR:
But I think it makes another issue at reconnect()'s if statement.
If I set max_retries to 3, bootstrap.connect() in reconnect() called 2 times, not 3 times.

1st reconnect(): tried_count = 0 => 1 (updated), if statement is 'true', retried
2nd reconnect(): tried_count = 1 => 2 (updated), if statement is 'true', retried
3rd reconnect(): tried_count = 2 => 3 (updated), if statement is 'false', closed

Do I understand flow correctly? Is it author's intention?

---

@d2r already pointed that issue to diff comment, while I see comment just now.

brndnmtthws:
I also misinterpreted that.

Regardless, there should be a hard limit of 30 bits for the shift operation, and 31 for max_retries.

anfeng:
@HeartSaVioR has a point here.

@brndnmtthws, how about that we make the following changes? 
  1) Change defaults.yaml to have 30 as max_retries
  2) Change reconnect()'s if statement to use "<=" instead of current "<"
  3) Revise getSleepTimeMs() as below:

```
private int getSleepTimeMs()
{
    int backoff = 1 << Math.max(1, retries.get());
    int sleepMs = base_sleep_ms * Math.max(1, random.nextInt(backoff));
    if ( sleepMs > max_sleep_ms )
        sleepMs = max_sleep_ms;
    return sleepMs;
}
```

HeartSaVioR:
@anfeng If you apply changes (2), (1)'s max_retries should be 30, not 31.
If we would apply (2), we should take care of getSleepTimeMs() to check any problem with retries == max_retries.
Shift operation and max_retries will having same hard limit, 30, as @brndnmtthws pointed.

anfeng:
@HeartSaVioR you are right on that. Let me just edit my suggestion to 30.

ptgoetz:
@brndnmtthws Are you okay with updating the pull request according to @anfeng's suggestion above?

@anfeng Can you review the pull request once it's updated?

This is the last issue that needs to be resolved before 0.9.0 is released. I'd like to get it merged soon.

brndnmtthws:
@ptgoetz Yeah, I'll update this in a bit.

brndnmtthws:
Updated as per the comments above.

Also did a whitespace cleanup.

ptgoetz:
+1 Thanks!

storm-netty/src/jvm/backtype/storm/messaging/netty/Client.java
```diff
@@ -38,14 +38,14 @@
...
+        message_queue = new LinkedBlockingQueue<Object>();
         retries = new AtomicInteger(0);
         channelRef = new AtomicReference<Channel>(null);
         being_closed = new AtomicBoolean(false);
 
-        // Configure 
+        // Configure
         buffer_size = Utils.getInt(storm_conf.get(Config.STORM_MESSAGING_NETTY_BUFFER_SIZE));
-        max_retries = Utils.getInt(storm_conf.get(Config.STORM_MESSAGING_NETTY_MAX_RETRIES));
+        max_retries = Math.max(30, Utils.getInt(storm_conf.get(Config.STORM_MESSAGING_NETTY_MAX_RETRIES)));
```
anfeng: This should be Math.min(). No?

brndnmtthws: Good catch! I get this wrong about 90% of the time :)

brndnmtthws:
Fixed the issue @anfeng pointed out with `Math.max` on `max_retries`.

anfeng:
Almost there. Can you remove the following statement from getSleepTimeMs()
        if ( sleepMs < base_sleep_ms )
          sleepMs = base_sleep_ms;
That should not be required w/ Math.max(1, ...) in line 97.

brndnmtthws:
@anfeng Another good catch.  Done.

anfeng:
+1 Thanks

ptgoetz:
Awesome. Thanks guys.

I'll cut the 0.9.0 release.

PR #731, opened 2013-11-07
Multiple output streams for trident

alienrobotwizard:
Addresses issue 638 (http://github.com/nathanmarz/storm/issues/638). @mridulj @mrflip

Consider:

``` java

Map<String, Fields> outputStreamMapping = new HashMap<String, Fields>();
outputStreamMapping.put("streamA", new Fields("field_in_a"));
outputStreamMapping.put("streamB", new Fields("field_in_b"));

Stream s = topology.newStream("testStream", spout);
StreamCollection multiStream = s.each(
  s.getOutputFields(),
  new DebugMultiStream("streamA", "streamB"),
  new StreamMap(outputStreamMapping)            
  );

Stream streamA = multiStream.getStream("streamA")
streamA.each(new Fields("field_in_a"), new Debug());
```

### DIFF
--- a/storm-core/src/jvm/storm/trident/Stream.java
+++ b/storm-core/src/jvm/storm/trident/Stream.java
@@ -1,11 +1,15 @@
 package storm.trident;
 
+import java.util.Map;
+import java.util.HashMap;
 import backtype.storm.generated.Grouping;
 import backtype.storm.generated.NullStruct;
 import storm.trident.fluent.ChainedAggregatorDeclarer;
 import backtype.storm.grouping.CustomStreamGrouping;
 import backtype.storm.tuple.Fields;
 import backtype.storm.utils.Utils;
+import storm.trident.fluent.StreamMap;
+import storm.trident.fluent.StreamCollection;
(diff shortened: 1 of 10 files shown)


nathanmarz:
I've only glanced at this, but I have two initial comments:
1. This needs tests
2. I'm not so sure about the API. I think that when the multistream each is used, it should return a "MultiStream" object instead of a stream object. And then the MultiStream object can be queried for the particular streams.

alienrobotwizard:
Nathan,
1. Yes, absolutely. That's the next thing. I just wanted to get the patch out asap since @mridulj was asking about it.
2. The multi each returns a StreamCollection object that has a 'getStream' method. It could just as easily be called MultiStream if that seems like a clearer name.

alienrobotwizard:
Oh I see the confusion. I've been writing my topologies in jruby which is why I can reuse 's' up there for both a Stream and a StreamCollection. I just did a quick conversion as I was typing. But my previous comment holds, it does return a collection in the way you're asking,

PR #733, opened 2013-11-08
Enable DRPC request to be sent via HTTP and/or Thrift

anfeng:
Currently, DRPC requests could only be sent via Thrift API. We have seen various users asking for HTTP interface.

This pull request enable one to configure drpc to be sent via HTTP and/or thrift via storm.yaml:
  drpc.port: <THRIFT_PORT> (default: 3772)
  drpc.http.port: <HTTP_PORT> (default: unavailable)

When drpc server is started, it will look into these configuration parameters to decide whether Thrift port and/or HTTP port should be binded.

DRPC HTTP request will be received via GET via the following URI:
- http://&lt;server>:&lt;HTTP_PORT>/drpc/&lt;Func>/&lt;Args>
- http://&lt;server>:&lt;HTTP_PORT>/drpc/&lt;Func>/
- http://&lt;server>:&lt;HTTP_PORT>/drpc/&lt;Func>


storm-core/src/clj/backtype/storm/daemon/drpc.clj
```diff
@@ -97,35 +101,63 @@
...
+    (handler request)))
+
+(defn webapp [handler]
+  (->(def http-routes
+      (routes
+        (GET "/drpc/:func/:args" [:as {cookies :cookies} func args & m]
+          (.execute handler func args))
+        (GET "/drpc/:func/" [:as {cookies :cookies} func & m]
+          (.execute handler func ""))
+        (GET "/drpc/:func" [:as {cookies :cookies} func & m]
```
revans2: I think this is in here twice by accident.

storm-core/src/clj/backtype/storm/daemon/drpc.clj
```diff
@@ -97,35 +101,63 @@
...
-          service-handler (service-handler)
+          drpc-http-port (if (conf DRPC-HTTP-PORT) (int (conf DRPC-HTTP-PORT)) 0)
+          drpc-port (int (conf DRPC-PORT))
+          drpc-service-handler (service-handler)
           ;; requests and returns need to be on separate thread pools, since calls to
           ;; "execute" don't unblock until other thrift methods are called. So if 
           ;; 64 threads are calling execute, the server won't accept the result
           ;; invocations that will unblock those threads
-          handler-server (THsHaServer. (-> (TNonblockingServerSocket. (int (conf DRPC-PORT)))
+          handler-server (if (> drpc-port 0)
```
revans2: We cannot turn off Thrift unless we provide HTTP APIs for the spout/return bolt, and update them to know how to use the HTTP APIs.

revans2:
It looks good.

storm-core/src/clj/backtype/storm/daemon/drpc.clj
anfeng: Thrift port will be turned off only if drpc server was launched with configuration stating so. As the default, it will be enabled.

storm-core/src/clj/backtype/storm/daemon/drpc.clj
revans2: Oops I missed the trailing '/'.  Looks like jetty is not handling '/' characters very well.

PR #740, opened 2013-11-11
Let kryo registration conf validation allow custom serializers.

revans2:
topology.kryo.register's validator did not allow for custom serializes, only the FieldSerializer.


revans2:
Already gone in, through a different channel

PR #746, opened 2013-11-14
deep hash code for byte[] and other array in tuple

xiaokang:
This PR is for #245 to fix the default hash code for byte[] and other array in tuple.


storm-core/src/clj/backtype/storm/tuple.clj
```diff
@@ -4,5 +4,33 @@
 
 (bootstrap)
 
+(def ARRAY-TYPES-MAP
```
nathanmarz: This should be a set. The syntax is #{}

storm-core/src/clj/backtype/storm/tuple.clj
```diff
@@ -4,5 +4,33 @@
...
+    (type (double-array 0)) nil
+    (type (boolean-array 0)) nil
+  })
+
+(defn deep-hash-code [alist]
+  "deep hash code based on array/list content.
+   it's the same as java.util.Arrays.deepHashCode(),
+   but without convert a list to array to avoid copy."
+  (reduce
+    #(.intValue (+ (.intValue (* 31 %1)) (hash-code %2)))
```
nathanmarz: How does this even compile? hash-code isn't declared anywhere

storm-core/src/clj/backtype/storm/tuple.clj
```diff
@@ -4,5 +4,33 @@
...
+    (type (float-array 0)) nil
+    (type (double-array 0)) nil
+    (type (boolean-array 0)) nil
+  })
+
+(defn deep-hash-code [alist]
+  "deep hash code based on array/list content.
+   it's the same as java.util.Arrays.deepHashCode(),
+   but without convert a list to array to avoid copy."
+  (reduce
```
nathanmarz, line 28: We can make this code much faster by not doing the reduce (which is constantly boxing and unboxing the numbers) and instead iterating through the list. It's doable in Clojure, though I'm ok with writing this code in Java and then calling it from here.

nathanmarz:
Made a few comments on the code.

storm-core/src/clj/backtype/storm/tuple.clj
xiaokang: Yes. It should be a set instead of map. It's due to my mis-understanding that contains? only work with map.

storm-core/src/clj/backtype/storm/tuple.clj
xiaokang: I am really sorry that the code cause compiling error and I have not tested it carefully. I just tested it in Clojure REPL using load-file. I'll fix it and test it on cluster.

storm-core/src/clj/backtype/storm/tuple.clj
xiaokang, line 28: Do you mean ".intValue" and "+" cause constantly boxing and unboxing the numbers? Can we use unchecked-add-int and unchecked-multiply-int to avoid boxing and unboxing? 
I tested with unchecked-\* and it's 10% faster than ".intValue" and "+" on 100KB and 1000KB byte array.

xiaokang:
Thanks @nathanmarz 's comments. The first and second is easy to fix but I am not very sure about the third.

nathanmarz:
I believe it will still box and unbox even when you use unchecked-add-int because it's wrapped in a function passed to reduce. I think it's worth trying to write that code in Java and see how the performance compares to the pure Clojure version. Alternatively, you can manually do a loop on the list – I believe Clojure will not box the bindings on the loop between iterations.

PR #749, opened 2013-11-18
Added in ability for users to include pre-installed jars to the classpath

revans2:


storm-core/src/clj/backtype/storm/util.clj
```diff
@@ -477,7 +477,7 @@
   (System/getProperty "java.class.path"))
 
 (defn add-to-classpath [classpath paths]
-  (str/join ":" (cons classpath paths)))
+  (str/join ":" (cons classpath (remove str/blank? paths))))
```
d2r, line 480: Can we use the `File/pathSeparator` here in place of `":"`?

d2r, line 480: Never mind, JVM classpath always uses a colon.

PR #755, opened 2013-11-26
Fix netty transport unit tests

ptgoetz:
Unit tests were failing because they did not define values for the number of client/server worker threads.


nathanmarz:
Should there just be defaults for those configs in defaults.yaml?

ptgoetz:
Values are there in `defaults.yaml`, but the netty transport tests use hard-coded config values rather than read from that file.

nathanmarz:
Oh I see. Cool, +1

xumingming:
+1

PR #756, opened 2013-11-27
Fix ClassCastException (was: logs a warning if we fail to delete a file)

d2r:
prevents ClassCastException


d2r:
Saw this on a file permission error.

We may want to throw a RuntimeException here instead?

revans2:
I don't think we want to throw a RuntimeException.  That would bring down nimbus, and because the file permissions were not fixed it would just crash again the next time it came up.  Because it is only cleaning up the files logging the error I think is enough.

nathanmarz:
-1. It's definitely an error and an indication that something is wrong with the setup.

d2r:
OK, so do we want to call log-error correctly then, since it expects a Throwable as the first argument (not a String)?

PR #766, opened 2013-12-10
Replace 0MQ with Netty

ptgoetz:
This is required in order to release from Apache. Because 0MQ/JZMQ is LGPL-licensed, it can't be included as a dependency in Apache releases.

This patch simply removes JZMQ/0MQ and replaces it with the netty transport implementation.

We can still make the 0MQ transport available to users who want to use it. But it will have to be a separate project that's not part of Apache. I will follow up with creating a repo that has the 0MQ transport code in isolation.


apetresc:
Can you give a brief breakdown of the tradeoffs being made? I'm assuming the change is backwards-compatible in terms of the API, but what are the functional differences between 0MQ and Netty as far as Storm is concerned? Should we expect better/worse performance on any axis?

jmlogan:
I know we have had issues running Netty that I have been unable to resolve so far...it's hitting some system resource limit, but 0MQ works fine.

Netty is supposed to perform significantly better than 0MQ (~5x iirc), and is pure-java, so it does not require any system installations (as 0MQ did). I'm not sure where Netty buffers things, but I would imagine buffers are dragged on-heap now, as opposed to 0MQ where they were off-heap, which could cause issues.

ptgoetz:
@apetresc @jmlogan For a good look at the performance characteristics of Netty vs. 0MQ, check out @revans2's [blog post](http://yahooeng.tumblr.com/post/64758709722/making-storm-fly-with-netty) about it.

As of storm 0.9.0, there are two transport options, Storm's original 0MQ transport and a new Netty-based transport. They are interchangeable via the storm configuration.

All this pull request does is remove the 0MQ transport (so it can move to its own separate project) and make the Netty transport the default. This is necessary since the 0MQ/JZMQ license (LGPL) is not compatible with the Apache license.

That being said, it will still be entirely possible to use the 0MQ transport, it just means that the source code and binaries could not be included in any official Apache release -- They would have to be downloaded and installed separately.

To illustrate, I put together a [project that isolates the 0MQ specific code](https://github.com/ptgoetz/storm-0mq) (there's really not much of it). To use the 0MQ transport all you would have to do is drop that project's jar file into the $STORM_HOME/lib directory, and add the following line to `storm.yaml`:

```
storm.messaging.transport: "backtype.storm.messaging.zmq"
```

Obviously you would also need the native binaries for 0MQ/JZMQ installed.

A side benefit of this change is that it would eliminate the need to install 0MQ/JZMQ in order to build Storm from source (a lot of people have run into issues doing so, myself included).

apetresc:
Thanks!

d2r:
Might it be nice to remove the unrelated changes (re-ordering, fixing up imports, etc.) to make the diff easier to read?

ptgoetz:
@d2r I apologize for that. It was not intentional.

It's largely a straight-up move, and change of config to use the netty transport. The only thing I had to change was to set the starting port number in `testing.clj` to 1024 so netty doesn't try to bind to privileged ports.

ptgoetz:
Link to corresponding JIRA:

https://issues.apache.org/jira/browse/STORM-1

revans2:
I ran into the resource limit issue too.  Essentially a thread is created in each worker to talk to every single other worker.  So you can easily hit the the thread ulimit.  For really large topologies performance will start to suffer because of the extra context switching involved.  

I know @anfeng took a quick look at this, but it is not a trivial change.

ptgoetz:
@revans2 Would you or @anfeng be able to open a JIRA issue for that?

https://issues.apache.org/jira/browse/STORM

revans2:
Sure I created https://issues.apache.org/jira/browse/STORM-12 based off of of what we have seen and the investigation we have done so far.

ptgoetz:
Perfect. Thanks @revans2!

ptgoetz:
Moved to Apache github project: https://github.com/apache/incubator-storm/pull/9

priyaswomi:
I am very new to apache storm. I am trying to run storm in remote mode. I don't know actually the steps to run in remote mode. Should i want to edit something in yaml file? should i want to mention the replacement of ZEROMQ with the NETTY. And also the steps to run in remote. Hope someone will help me.

PR #768, opened 2013-12-11
Update Dependencies to Released Versions 

ptgoetz:
This addresses the following JIRA issues and moves us closer to being able to release from Apache:

https://issues.apache.org/jira/browse/STORM-2

https://issues.apache.org/jira/browse/STORM-6

**jgrapht**
The reason for the bump of jgrapht to 0.9.0 is its license. Version 0.8.3 was LGPL, while 0.9.0 switched to a dual EPL/LGPL license. The choice here was either get rid of jgrapht and find an Apache compatible replacement (Hard), or upgrade to 0.9.0 (Easy).

**carbonite**
This took some digging. It looks like the only modifications Nathan's version were to align the version of kryo with the version that Storm used. Verstion 1.3.2 from `com.twitter` depends on the same version of kryo as Storm.

**tools.cli**
I couldn't find any indication that the "storm" version differed from the "org.clojure" version.

With these modifications, all unit tests pass. I also tested on a 3-node cluster by running a standard topology, a trident topology, and DRPC. Everything checked out okay.


nathanmarz:
+1. I think the tools.cli dependency was made because it wasn't available in a Maven repo at the time, so I just put it in myself under the "storm" group.

sritchie:
Just to note, this is now pegged to my fork of Carbonite at https://github.com/sritchie/carbonite. Cascalog depends on this version as well. @johnynek, @jasonjckn, @nathanmarz - would a good next step be migrating my fork into Chill?

PR #769, opened 2013-12-12
Use Stock libthrift dependency

ptgoetz:
This is hopefully a temporary workaround for the fact that storm has been using a non-stock build of the `libthrift` dependency. Hopefully it will spur discussions that lead to a better solution.

It works by downloading the stock `libthrift` jar, uses jarjar to rewrite package names, then installs the resulting jar in the local maven repository where leiningen and maven will be able to find it.

One downside is that users would have to run ./bin/install_libthrift.sh (only once) before building from source. So it kind of muddies the build process.

Ideally I'd like to see this functionality integrated with the leiningen build process so we don't rely on one-off shell scripts to build.

### DIFF
--- a/bin/build_release.sh
+++ b/bin/build_release.sh
@@ -18,6 +18,7 @@ rm -f *.tar.gz
 $LEIN pom || exit 1
 mkdir -p $DIR/lib
 
+sh bin/install_libthrift7.sh
 
 sh bin/build_modules.sh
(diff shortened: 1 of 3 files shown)


ptgoetz:
Moved to apache/incubator-storm:

https://github.com/apache/incubator-storm/pull/7

PR #781, opened 2014-01-20
Added logging level to multilang protocol spout and bolt.

mahall:
I would like to be able to log at varying levels from multilang spouts and bolts. I updated the log function in storm.py to make the call.

I propose the following change in the wiki doc https://github.com/nathanmarz/storm/wiki/Multilang-protocol as well.

{
    "command": "log",
    // the message to log
    "msg": "hello world!",
    // the level to log at (optional, defaults to info)
    "level": "warn"
}


ptgoetz:
Storm has moved to https://github.com/apache/incubator-storm

Please redirect your pull request to that repo.

